PR #4, opened 2018-05-01
[141368] Quit using macro to deliver default mix configurations; use ordinary function instead

skirino:
https://acsmine.tok.access-company.com/redmine/issues/141368


ymtszw:
# okay

CI failure looks unrelated but not sure what is the cause...

PR #5, opened 2018-05-21
143354_setup_project_config_for_hex_publish

ymtszw:
https://acsmine.tok.access-company.com/redmine/issues/143354


lib/type/string.ex
```diff
@@ -7,6 +7,10 @@ defmodule Antikythera.VersionStr do
   @moduledoc """
   Format of versions of antikythera instances and gears.
 
+  Mostly generated by using `Antikythera.MixCommon.version_with_last_commit_info/1` defined in `mix_common.exs` file.
+  It forces incrementing OTP application versions of antikythera instances or gears according to current time and git commit hash.
+  Useful if you are going to continuously deploy them on every commit to target branches.
```
skirino: Explaining the rationale of the version string format here is a good idea, but we should make things more accurate.
e.g. we don't use current time (if so it'd always make a new version just by recompiling); we use committer date.
Try this:

```
The format rule is a stricter variant of [semantic versioning](https://semver.org/);
pre-release part and build metadata part are filled with information from the current git commit (committer date and commit hash, respectively).
This way we always make a new version for each new git commit, which enables both antikythera instances and gears to be deployed on a per-commit basis.
The actual version strings are created by `Antikythera.MixCommon.version_with_last_commit_info/1` defined in `mix_common.exs` file.
```

mix.exs
```diff
@@ -120,6 +122,18 @@ defmodule Antikythera.Mixfile do
     ]
   end
 
+  defp package() do
+    [
+      licesnses:   ["Apache 2.0"],
+      maintainers: ["antikythera-gr@access-company.com"],
+      links:       %{"GitHub" => @github_url},
+      files:       [
+        "assets", "core", "eal", "lib", "local", "priv", "rel",
```
skirino: The png file in `"assets"` dir looks unnecessary to me. Can we omit `"assets"` from the hex package?

mix.exs
```diff
@@ -36,12 +36,14 @@ defmodule Antikythera.Mixfile do
   def project() do
     [
       app:             :antikythera,
-      version:         Antikythera.MixCommon.version_with_last_commit_info("0.1.0"),
+      version:         "0.1.0",
```
skirino: Versions such as `0.1.0-20180511053513+67ba650c7885eee89ec672d147f1d2c469916f6d` already exist as git commits, although they are not available as hex package.
Should we start with `0.1.1` or `0.2.0`, so that it is strictly newer than all existing versions?
(in this sense, I shouldn't have modified the version from `0.0.1` to `0.1.0` a few weeks ago...)

lib/type/string.ex
ymtszw: Looking good. de8226a
Current time thing was a foolish error on my part..

test/lib/type/string_test.exs
```diff
@@ -5,8 +5,10 @@ defmodule Antikythera.StringTypesTest do
   alias Antikythera.{VersionStr, Domain, EncodedPath, UnencodedPath, Email, Url, ContextId, TenantId}
 
   test "validate VersionStr" do
-    v = Mix.Project.config()[:version]
-    assert VersionStr.valid?(v)
+    assert VersionStr.valid?("0.0.1-20180501235959+0123456789abcdef0123456789abcdef01234567")
+
+    # multi-digit numbers in semantic version part are currently not allowed
```
skirino: The whole version string can be seen as a "semantic version part".
We should write something like
"multi-digit numbers in major/minor/patch part are currently not allowed"

skirino:
I've just made another comment.
LGTM, no re-review required.

test/lib/type/string_test.exs
ymtszw: Fixed in d2bd470
Also showing a valid `Antikythera.VersionStr` string is parsable by Elixir's `Version` too.

test/lib/type/string_test.exs
```diff
@@ -5,9 +5,14 @@ defmodule Antikythera.StringTypesTest do
...
 
   test "validate VersionStr" do
-    assert VersionStr.valid?("0.0.1-20180501235959+0123456789abcdef0123456789abcdef01234567")
+    valid_version_str = "0.0.1-20180501235959+0123456789abcdef0123456789abcdef01234567"
+    assert VersionStr.valid?(valid_version_str)
 
-    # multi-digit numbers in semantic version part are currently not allowed
+    # It is also valid as a standard semantic version string
+    {:ok, v} = Version.parse(valid_version_str)
+    assert v == %Version{major: 0, minor: 0, patch: 1, pre: [20180501235959], build: "0123456789abcdef0123456789abcdef01234567"}
```
ymtszw: Credo barks on `20180501235959`, demands splitting with `_`.
Though it is constructed from date-time so I purposefully kept it un-split.
In semantic version, [all-numeric pre-release parts are considered as numbers and compared numerically](https://semver.org/#spec-item-11).

PR #7, opened 2018-05-22
143354_change_path_to_documentation_sources

ymtszw:
https://acsmine.tok.access-company.com/redmine/issues/143354


skirino:
I'm not a big fan of the name `guides/`, as It's not obvious where to put non-guide docs such as internal design docs.
A possible candidate I can think of is `doc_src`;
it implies that exdoc-generated HTMLs are canonical, markdown files are input for exdoc.

ymtszw:
> A possible candidate I can think of is doc_src;
> it implies that exdoc-generated HTMLs are canonical, markdown files are input for exdoc.

Cool! Let's go with that a2295be

mix.exs
```diff
@@ -44,7 +44,7 @@ defmodule Antikythera.Mixfile do
       homepage_url:    @github_url,
       description:     "An Elixir framework to build your own in-house PaaS (Platform as a Service).",
       package:         package(),
-    ] ++ Keyword.update!(Antikythera.MixCommon.common_project_settings(), :docs, &(&1 ++ docs()))
+    ] ++ Keyword.replace!(Antikythera.MixCommon.common_project_settings(), :docs, docs())
```
skirino: We should avoid overriding "common" settings here.
Now that a9a and gears don't share any `:docs` setting, let's remove `:docs` key in `Antikythera.MixCommon.common_project_settings()` and move
`docs: [output: "exdoc"]`
into both `Antikythera.GearProject` and our antikythera instance's setting.
After that we can simply write as follows:
```ex
[
  ...
  docs: docs(),
  ...
] ++ Antikythera.MixCommon.common_project_settings()

skirino:
I've just made another comment (no re-review required).
Other than that, LGTM

PR #8, opened 2018-05-29
143632 rate limit queries and commands to async job queues

skirino:
https://acsmine.tok.access-company.com/redmine/issues/143632

### DIFF
--- a/test/core/async_job/queue_test.exs
+++ b/test/core/async_job/queue_test.exs
@@ -197,22 +197,17 @@ defmodule AntikytheraCore.AsyncJob.QueueTest do
     assert Queue.cancel(@queue_name, @job_id) == {:error, :not_found}
   end
 
-  test "consecutive registrations should be rejected by rate limit" do
+  test "consecutive registrations/cancels should be rejected by rate limit" do
     n = div(RateLimit.max_tokens(), RateLimit.tokens_per_command())
-    Enum.each(1..n, fn _ ->
-      assert register_job([]) == :ok
-    end)
-    {:error, {:rate_limit_reached, _}} = register_job([])
-  end
-
-  test "consecutive cancels should be rejected by rate limit" do
-    n = div(RateLimit.max_tokens(), RateLimit.tokens_per_command())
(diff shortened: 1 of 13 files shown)


test/core/async_job/queue_test.exs
```diff
@@ -159,35 +164,81 @@ defmodule AntikytheraCore.AsyncJob.QueueTest do
...
+    Enum.each(1..n, fn _ ->
+      assert register_job([]) == :ok
+    end)
+    {:error, {:rate_limit_reached, _}} = register_job([])
+  end
+
+  test "consecutive cancels should be rejected by rate limit" do
+    n = div(RateLimit.max_tokens(), RateLimit.tokens_per_command())
+    Enum.each(0..n, fn i ->
+      AsyncJobHelper.reset_rate_limit_status(@epool_id)
```
ymtszw: Any intent behind calling `AsyncJobHelper.reset_rate_limit_status/1` on every iteration?
Since it looks to me that we can do this:
```ex
Enum.each(1..n, fn i ->
  assert register_job([job_id: @job_id <> "#{i}"]) == :ok
end)
AsyncJobHelper.reset_rate_limit_status(@epool_id)
assert register_job([job_id: @job_id <> "0"]) == :ok
Enum.each(1..n, fn i ->
  assert Queue.cancel(@queue_name, @job_id <> "#{i}")
end)
{:error, {:rate_limit_reached, _}} = Queue.cancel(@queue_name, @job_id <> "0")
```

skirino: I've fixed some of my silly mistakes (missing `== :ok` and wrong key name `:job_id` => `:id`).
(Thank you for making me aware of them)
I'm not particularly concerned about how we "reset" the token bucket; as you showed less function call is better.
And actually we can make the tests a bit more compact by sharing the test setup for job cancellation and assertions for job registration.
https://github.com/access-company/antikythera/pull/8/commits/48b65c2543bf3d37737e79d94b3edd1c62d6898e

ymtszw:
# okay

PR #10, opened 2018-06-08
[133935] Cache gear config in process dictionary while handling web requests

s417-lama:
Ticket: https://acsmine.tok.access-company.com/redmine/issues/133935

### DIFF
--- a/lib/gear_application/config_getter.ex
+++ b/lib/gear_application/config_getter.ex
@@ -27,6 +27,32 @@ defmodule Antikythera.GearApplication.ConfigGetter do
 
   @doc false
   def get_all_env(gear_name) do
+    case load_config_from_process_dictionary(gear_name) do
+      {:ok   , config              } -> config
+      {:error, current_gear_configs} ->
+        config = load_config_from_ets(gear_name)
+        store_config_to_process_dictionary(gear_name, current_gear_configs, config)
+        config
+    end
+  end
+
+  defp load_config_from_process_dictionary(gear_name) do
+    case Process.get(:gear_configs) do
+      nil     -> {:error, %{}}
+      configs ->
+        case Map.get(configs, gear_name) do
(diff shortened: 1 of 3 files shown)


lib/gear_application/config_getter.ex
```diff
@@ -27,6 +27,32 @@ defmodule Antikythera.GearApplication.ConfigGetter do
...
+      {:ok   , config              } -> config
+      {:error, current_gear_configs} ->
+        config = load_config_from_ets(gear_name)
+        store_config_to_process_dictionary(gear_name, current_gear_configs, config)
+        config
+    end
+  end
+
+  defp load_config_from_process_dictionary(gear_name) do
+    case Process.get(:gear_configs) do
```
skirino: Let's make this atom a module attribute.
```ex
@key :gear_configs
```

lib/gear_application/config_getter.ex
```diff
@@ -27,6 +27,32 @@ defmodule Antikythera.GearApplication.ConfigGetter do
...
+      nil     -> {:error, %{}}
+      configs ->
```
skirino: `config` and `configs` can be a bit confusing. Let's use `gear_configs` for this variable.
Also let's rename `current_gear_configs` to `gear_configs` (just to make it shorter and consistent).

skirino:
Made 2 comments, no re-review required.

lib/gear_application/config_getter.ex
s417-lama: fixed in 40400f1f530816cdb00b656fc408cec823d826e1

lib/gear_application/config_getter.ex
s417-lama: fixed in 40400f1f530816cdb00b656fc408cec823d826e1

ymtszw:
# okay

PR #12, opened 2018-06-14
[146113] Implement handle_cast in GearConfigPoller

ttokit:
https://acsmine.tok.access-company.com/redmine/issues/146113

### DIFF
--- a/core/sup_tree_core/gear_config_poller.ex
+++ b/core/sup_tree_core/gear_config_poller.ex
@@ -1,5 +1,7 @@
 # Copyright(c) 2015-2018 ACCESS CO., LTD. All rights reserved.
 
+use Croma
+
 defmodule AntikytheraCore.GearConfigPoller do
   @moduledoc """
   Periodically polls changes in gear configs.
@@ -20,6 +22,8 @@ defmodule AntikytheraCore.GearConfigPoller do
 
   @interval 60_000
 
+  @typep state :: %{last_checked_at: integer}
+
   def start_link() do
     GenServer.start_link(__MODULE__, :ok, [name: __MODULE__])
   end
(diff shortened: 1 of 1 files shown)


core/sup_tree_core/gear_config_poller.ex
```diff
@@ -29,6 +29,13 @@ defmodule AntikytheraCore.GearConfigPoller do
     {:ok, %{last_checked_at: 0}, @interval}
   end
 
+  @impl true
+  def handle_cast(:reload, state) do
+    checked_at = System.system_time(:seconds)
+    GearConfig.load_all(state[:last_checked_at])
+    {:noreply, %{state | last_checked_at: checked_at}, @interval}
```
skirino: Don't copy existing code.
Let's make a private function to extract the common part.

ttokit: fixed in 75f01f2

core/sup_tree_core/gear_config_poller.ex
```diff
@@ -20,6 +22,8 @@ defmodule AntikytheraCore.GearConfigPoller do
 
   @interval 60_000
 
+  @typep state :: %{last_checked_at: integer}
```
skirino: Declaring this as `@typep` is :+1:
We can make the value type more specific by `integer` => `pos_integer`, as seconds since UNIX epoch is always positive.

core/sup_tree_core/gear_config_poller.ex
```diff
@@ -29,10 +33,21 @@ defmodule AntikytheraCore.GearConfigPoller do
     {:ok, %{last_checked_at: 0}, @interval}
   end
 
+  @impl true
+  def handle_cast(:reload, state) do
+    new_state = load_gear_config(state)
+    {:noreply, new_state, @interval}
```
skirino: "Making a 3-tuple" is also duplicated for the 2 handler callbacks.
Let's add this part in the common private function (and rename it).

core/sup_tree_core/gear_config_poller.ex
ttokit: fixed in ba40bc0

ymtszw:
# okay

core/sup_tree_core/gear_config_poller.ex
```diff
@@ -29,8 +33,17 @@ defmodule AntikytheraCore.GearConfigPoller do
...
+  def handle_cast(:reload, state) do
+    handle_gear_config_loading(state)
+  end
+
   @impl true
   def handle_info(:timeout, state) do
+    handle_gear_config_loading(state)
+  end
+
+  defunp handle_gear_config_loading(state :: state_t) :: state_t do
```
skirino: Now the return type isn't `state_t`; it should be `{:noreply, state_t, timeout}`.

skirino:
1 more comment, proceed when it's done!

core/sup_tree_core/gear_config_poller.ex
ttokit: Oops, I had overlooked.
fixed in 9a7571c

PR #13, opened 2018-06-18
[145135] Disable gear upgrades when a node recognizes that it's about to be terminated

ttokit:
- Add function to enable / disable upgrade of antikythera instance or gears in VersionUpgradeTaskQueue

https://acsmine.tok.access-company.com/redmine/issues/145135

### DIFF
--- a/core/sup_tree_core/version_upgrade_task_queue.ex
+++ b/core/sup_tree_core/version_upgrade_task_queue.ex
@@ -77,18 +77,18 @@ defmodule AntikytheraCore.VersionUpgradeTaskQueue do
   # Public API
   #
   defun gear_updated(gear_name :: v[GearName.t]) :: :ok do
-    :ok = GenServer.cast(__MODULE__, {:gear_updated, gear_name})
+    GenServer.cast(__MODULE__, {:gear_updated, gear_name})
   end
 
   defun core_updated() :: :ok do
-    :ok = GenServer.cast(__MODULE__, :core_updated)
+    GenServer.cast(__MODULE__, :core_updated)
   end
 
-  defun enable_version_upgrade() :: :ok do
-    :ok = GenServer.cast(__MODULE__, :enable_version_upgrade)
+  defun enable() :: :ok do
+    GenServer.cast(__MODULE__, :enable_version_upgrade)
   end
 
-  defun disable_version_upgrade() :: :ok do
-    :ok = GenServer.cast(__MODULE__, :disable_version_upgrade)
+  defun disable() :: :ok do
+    GenServer.cast(__MODULE__, :disable_version_upgrade)
   end
 end
--- a/core/sup_tree_core/version_upgrade_task_queue.ex
+++ b/core/sup_tree_core/version_upgrade_task_queue.ex
@@ -65,11 +65,11 @@ defmodule AntikytheraCore.VersionUpgradeTaskQueue do
     end
   end
 
-  defp run_task({:gear_updated, gear_name}) do
+  defp run_task({:gear, gear_name}) do
     {pid, _ref} = spawn_monitor(Version.Gear, :install_or_upgrade_to_next_version, [gear_name])
     pid
   end
-  defp run_task(:core_updated) do
+  defp run_task(:core) do
     {pid, _ref} = spawn_monitor(Version.Core, :upgrade_to_next_version, [])
     pid
   end
@@ -78,11 +78,11 @@ defmodule AntikytheraCore.VersionUpgradeTaskQueue do
   # Public API
   #
   defun gear_updated(gear_name :: v[GearName.t]) :: :ok do
-    GenServer.cast(__MODULE__, {:upgrade, {:gear_updated, gear_name}})
+    GenServer.cast(__MODULE__, {:upgrade, {:gear, gear_name}})
   end
 
   defun core_updated() :: :ok do
-    GenServer.cast(__MODULE__, {:upgrade, :core_updated})
+    GenServer.cast(__MODULE__, {:upgrade, :core})
   end
 
   defun enable() :: :ok do
--- a/core/sup_tree_core/version_upgrade_task_queue.ex
+++ b/core/sup_tree_core/version_upgrade_task_queue.ex
@@ -58,7 +58,7 @@ defmodule AntikytheraCore.VersionUpgradeTaskQueue do
     if pid == nil and enabled? do
       case :queue.out(q) do
         {{:value, instruction}, new_queue} -> %{state | queue: new_queue, task_pid: run_task(instruction)}
-        {:empty           , _        } -> state
+        {:empty               , _        } -> state
       end
     else
       state
--- a/core/sup_tree_core/version_upgrade_task_queue.ex
+++ b/core/sup_tree_core/version_upgrade_task_queue.ex
@@ -54,7 +54,7 @@ defmodule AntikytheraCore.VersionUpgradeTaskQueue do
     {:noreply, new_state}
   end
 
-  defp run_task_if_possible(%{queue: q, task_pid: pid, enable_upgrade?: enable?} = state) do
+  defp run_task_if_possible(%{queue: q, task_pid: pid, enable?: enable?} = state) do
     if pid == nil and enable? do
       case :queue.out(q) do
         {{:value, message}, new_queue} -> %{state | queue: new_queue, task_pid: run_task(message)}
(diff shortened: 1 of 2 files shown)


core/sup_tree_core/version_upgrade_task_queue.ex
```diff
@@ -70,4 +83,12 @@ defmodule AntikytheraCore.VersionUpgradeTaskQueue do
   defun core_updated() :: :ok do
     :ok = GenServer.cast(__MODULE__, :core_updated)
   end
+
+  defun enable_version_upgrade() :: :ok do
+    :ok = GenServer.cast(__MODULE__, :enable_version_upgrade)
+  end
+
+  defun disable_version_upgrade() :: :ok do
+    :ok = GenServer.cast(__MODULE__, :disable_version_upgrade)
```
skirino: - It's an issue about existing code, but `:ok = ` isn't necessary as `GenServer.cast/2` always returns `:ok`.
Let's remove them (4 lines).
- Names of these new functions are a bit redundant given that the module name contains "version" and "upgrade".
`enable/0` and `disable/0` should suffice.

core/sup_tree_core/version_upgrade_task_queue.ex
```diff
@@ -14,17 +14,30 @@ defmodule AntikytheraCore.VersionUpgradeTaskQueue do
...
+  def handle_cast(message, state) when message in [:enable_version_upgrade, :disable_version_upgrade] do
+    new_state =
+      case message do
+        :enable_version_upgrade  ->
+          L.info("All upgrade processing of antikythera instance or gears is set again to enable.")
+          %{state | enable_upgrade?: true}
+        :disable_version_upgrade ->
+          L.info("All upgrade processing of antikythera instance or gears is set to disable.")
+          %{state | enable_upgrade?: false}
+      end
```
skirino: Try something like the following:
```ex
def handle_cast(message, state) do
  new_state =
    case message do
      {:upgrade, instruction}  -> enqueue_message(state, instruction) |> run_task_if_possible()
      {:set_enabled, enabled?} -> set_enabled(state, enabled?)
    end
  {:noreply, new_state}
end

defp set_enabled(state, enabled?) do
  case {state.enabled?, enabled?} do
    {true , false} -> L.info("upgrade processing is turned off")
    {false, true } -> L.info("upgrade processing is turned on")
    _              -> :ok
  end
  %{state | enabled?: enabled?}
end
```

- Field name in process state should be `:enabled?` or `:upgrade_enabled?` rather than `:enable_upgrade?`.
- Now this `GenServer` receives 4 types of cast messages. Let's introduce a "structure" to classify them.

ttokit: It looks like a very smart code!

BTW, I came up with the implementation of "structure" below.
```
defmodule Message do
  use Croma.Struct, recursive_new?: true, fields: [
    upgrade:     Croma.TypeGen.nilable(Croma.Any),
    set_enabled: Croma.TypeGen.nilable(Croma.Boolean),
  ]
end
```
Is this data structure appropriate? Or is there a more appropriate data structure?

skirino: Here what we want is a "tagged union" (or variant).
In Erlang/Elixir world the simplest way to implement "tagged union" is to use tuple.

I'm not a big fan of a struct with 2 nilable fields here, because e.g.
`%Message{upgrade: nil, set_enabled: nil}` is treated as a valid value for the type.
In order to circumvent the issue we could override `valid?/1` and `new/1`, but it's not worth the effort.

Note that the following things are basically orthogonal to the choice of data structure:
- whether to declare the type or not
- whether to validate arguments (using `defun` family) or not

So here (I think) simply using tuples is the way to go.

core/sup_tree_core/version_upgrade_task_queue.ex
ttokit: fixed in 2139515

core/sup_tree_core/version_upgrade_task_queue.ex
ttokit: Oh, I had a big misunderstanding.
I thought tuple and "structure" are different things. .
Thank you.

fixed in 32812e9

core/sup_tree_core/version_upgrade_task_queue.ex
```diff
@@ -14,36 +14,50 @@ defmodule AntikytheraCore.VersionUpgradeTaskQueue do
...
+
   @impl true
   def handle_info({:DOWN, _monitor_ref, :process, pid, _reason}, %{task_pid: pid} = state) do
     new_state = Map.put(state, :task_pid, nil) |> run_task_if_possible()
     {:noreply, new_state}
   end
 
-  defp run_task_if_possible(%{queue: q, task_pid: pid} = state) do
-    if pid == nil do
+  defp run_task_if_possible(%{queue: q, task_pid: pid, enable_upgrade?: enable?} = state) do
```
skirino: `enable_upgrade?` here should also be renamed.

core/sup_tree_core/version_upgrade_task_queue.ex
```diff
@@ -64,10 +78,18 @@ defmodule AntikytheraCore.VersionUpgradeTaskQueue do
   # Public API
   #
   defun gear_updated(gear_name :: v[GearName.t]) :: :ok do
-    :ok = GenServer.cast(__MODULE__, {:gear_updated, gear_name})
+    GenServer.cast(__MODULE__, {:upgrade, {:gear_updated, gear_name}})
```
skirino: Now that `{:upgrade, _}` is in place, let's remove redundancy in the inner (original) atom:

- `{:upgrade, {:gear, gear_name}}`
- `{:upgrade, :core}`

should suffice.

core/sup_tree_core/version_upgrade_task_queue.ex
ttokit: Sorry, I'd overlooked.
fixed in e29d372

core/sup_tree_core/version_upgrade_task_queue.ex
ttokit: fixed in 9492230

core/sup_tree_core/version_upgrade_task_queue.ex
```diff
@@ -14,48 +14,62 @@ defmodule AntikytheraCore.VersionUpgradeTaskQueue do
...
+
   @impl true
   def handle_info({:DOWN, _monitor_ref, :process, pid, _reason}, %{task_pid: pid} = state) do
     new_state = Map.put(state, :task_pid, nil) |> run_task_if_possible()
     {:noreply, new_state}
   end
 
-  defp run_task_if_possible(%{queue: q, task_pid: pid} = state) do
-    if pid == nil do
+  defp run_task_if_possible(%{queue: q, task_pid: pid, enable?: enable?} = state) do
```
skirino: "d" is missing here.
(Tests on CircleCI is failing due to this)

core/sup_tree_core/version_upgrade_task_queue.ex
```diff
@@ -14,48 +14,62 @@ defmodule AntikytheraCore.VersionUpgradeTaskQueue do
...
+    if pid == nil and enable? do
       case :queue.out(q) do
-        {{:value, message}, new_queue} -> %{queue: new_queue, task_pid: run_task(message)}
+        {{:value, message}, new_queue} -> %{state | queue: new_queue, task_pid: run_task(message)}
```
skirino: Let's rename `message` to `instruction`.

core/sup_tree_core/version_upgrade_task_queue.ex
ttokit: fixed in baa9e5a

core/sup_tree_core/version_upgrade_task_queue.ex
```diff
@@ -14,48 +14,62 @@ defmodule AntikytheraCore.VersionUpgradeTaskQueue do
...
+    new_state =
+      case message do
+        {:upgrade    , instruction} -> enqueue_message(state, instruction) |> run_task_if_possible()
+        {:set_enabled, enabled?   } -> set_enabled(state, enabled?)
+      end
     {:noreply, new_state}
   end
 
   defp enqueue_message(%{queue: q} = state, message) do
     %{state | queue: :queue.in(message, q)}
```
skirino: Now we have the word "instruction" to represent the specific notion in this `GenServer`, let's make identifiers consistent with the term:
- function name: `enqueue_message` => `enqueue_instruction`
- parameter name: `message` => `instruction`

core/sup_tree_core/version_upgrade_task_queue.ex
ttokit: fixed in 8ed60cd, d028431

skirino:
A few more renaming.
When you are done it's OK for me.

ymtszw:
# okay 🚀

PR #14, opened 2018-06-20
[133936] Add Antikythera.Memcache as an in-memory cache storage

s417-lama:
- Store objects in ETS per epool_id
- Limit the number of objects and its each size
- Objects will be expired probabilistically (to avoid the thundering herd)

https://acsmine.tok.access-company.com/redmine/issues/133936

### DIFF
--- a/core/ets/memcache.ex
+++ b/core/ets/memcache.ex
@@ -4,7 +4,6 @@ use Croma
 
 defmodule AntikytheraCore.Ets.Memcache do
   alias Croma.Result, as: R
-  alias Antikythera.Time
   alias Antikythera.ExecutorPool.Id, as: EPoolId
 
   defun init() :: :ok do
@@ -22,7 +21,7 @@ defmodule AntikytheraCore.Ets.Memcache do
     end
   end
 
-  defun write(key :: term, value :: term, expire_at :: v[Time.t], prob_expire_at :: v[Time.t], epool_id :: v[EPoolId.t]) :: :ok do
+  defun write(key :: term, value :: term, expire_at :: integer, prob_expire_at :: integer, epool_id :: v[EPoolId.t]) :: :ok do
     :ets.insert(table_name(), {{epool_id, key}, expire_at, prob_expire_at, value})
     :ok
   end
--- a/core/ets/memcache.ex
+++ b/core/ets/memcache.ex
@@ -6,28 +6,27 @@ defmodule AntikytheraCore.Ets.Memcache do
   alias Croma.Result, as: R
   alias Antikythera.ExecutorPool.Id, as: EPoolId
 
-  defun init() :: :ok do
-    AntikytheraCore.Ets.create_read_optimized_table(table_name())
-  end
+  @table_name :antikythera_memcache
+  defun table_name() :: atom, do: @table_name
 
-  defun table_name() :: atom do
-    :antikythera_memcache
+  defun init() :: :ok do
+    AntikytheraCore.Ets.create_read_optimized_table(@table_name)
   end
 
   defun read(key :: term, epool_id :: v[EPoolId.t]) :: R.t(term, :not_found) do
-    case :ets.lookup(table_name(), {epool_id, key}) do
+    case :ets.lookup(@table_name, {epool_id, key}) do
       []        -> {:error, :not_found}
       [element] -> {:ok   , element}
     end
   end
 
   defun write(key :: term, value :: term, expire_at :: integer, prob_expire_at :: integer, epool_id :: v[EPoolId.t]) :: :ok do
-    :ets.insert(table_name(), {{epool_id, key}, expire_at, prob_expire_at, value})
+    :ets.insert(@table_name, {{epool_id, key}, expire_at, prob_expire_at, value})
     :ok
   end
 
   defun delete(key :: term, epool_id :: v[EPoolId.t]) :: :ok do
-    :ets.delete(table_name(), {epool_id, key})
+    :ets.delete(@table_name, {epool_id, key})
     :ok
   end
 end
--- a/core/ets/memcache.ex
+++ b/core/ets/memcache.ex
@@ -1,0 +1,34 @@
+# Copyright(c) 2015-2018 ACCESS CO., LTD. All rights reserved.
+
+use Croma
+
+defmodule AntikytheraCore.Ets.Memcache do
+  alias Croma.Result, as: R
+  alias Antikythera.Time
+  alias Antikythera.ExecutorPool.Id, as: EPoolId
+
+  defun init() :: :ok do
+    AntikytheraCore.Ets.create_read_optimized_table(table_name())
+  end
+
+  defun table_name() :: atom do
+    :antikythera_memcache
+  end
+
+  defun read(key :: term, epool_id :: v[EPoolId.t]) :: R.t(term, :not_found) do
+    case :ets.lookup(table_name(), {epool_id, key}) do
+      []        -> {:error, :not_found}
+      [element] -> {:ok   , element}
+    end
+  end
+
+  defun write(key :: term, value :: term, expire_at :: v[Time.t], prob_expire_at :: v[Time.t], epool_id :: v[EPoolId.t]) :: :ok do
+    :ets.insert(table_name(), {{epool_id, key}, expire_at, prob_expire_at, value})
+    :ok
+  end
+
+  defun delete(key :: term, epool_id :: v[EPoolId.t]) :: :ok do
+    :ets.delete(table_name(), {epool_id, key})
+    :ok
+  end
+end
--- a/core/sup_tree_core/executor_pool/memcache_manager.ex
+++ b/core/sup_tree_core/executor_pool/memcache_manager.ex
@@ -1,0 +1,104 @@
+# Copyright(c) 2015-2018 ACCESS CO., LTD. All rights reserved.
+
+use Croma
+
+defmodule AntikytheraCore.ExecutorPool.MemcacheManager do
+  @moduledoc """
+  A `GenServer` to manage memcache for each executor pool.
+  """
+
+  @timeout         300_000
+  @max_records_num 10
+  @max_record_size 65536
+
+  use GenServer
+  alias Antikythera.Time
+  alias AntikytheraCore.Ets.Memcache
+  alias AntikytheraCore.ExecutorPool.RegisteredName, as: RegName
+  alias Antikythera.ExecutorPool.Id, as: EPoolId
+
+  def start_link(name, epool_id) do
+    GenServer.start_link(__MODULE__, epool_id, [name: name])
+  end
+
+  @impl true
+  def init(epool_id) do
+    {:ok, %{epool_id: epool_id, records: %{}, expire_at_set: :gb_sets.new()}, @timeout}
+  end
+
+  @impl true
+  def handle_call({:write, key, value, lifetime, ratio}, _from, state) do
+    evicted_state = evict_expired_records(state, Time.now())
+    if :erts_debug.flat_size(value) > @max_record_size do
+      {:reply, {:error, :too_large_object}, evicted_state, @timeout}
+    else
+      new_state =
+        evicted_state
+        |> evict_head_record_if_exceeds_limit()
+        |> do_write(key, value, lifetime, ratio)
+      {:reply, :ok, new_state, @timeout}
+    end
+  end
+
+  @impl true
+  def handle_info(:timeout, state) do
+    new_state = evict_expired_records(state, Time.now())
+    {:noreply, new_state, @timeout}
+  end
+
+  defp evict_expired_records(%{records: records, expire_at_set: expire_at_set, epool_id: epool_id} = state, current_time) do
+    if :gb_sets.is_empty(expire_at_set) do
+      state
+    else
+      {{expire_at, key}, new_expire_at_set} = :gb_sets.take_largest(expire_at_set)
+      if expire_at < current_time do
+        Memcache.delete(key, epool_id)
+        new_records = Map.delete(records, key)
+        evict_expired_records(%{state | records: new_records, expire_at_set: new_expire_at_set}, current_time)
+      else
+        state
+      end
+    end
+  end
+
+  defp evict_head_record_if_exceeds_limit(%{records: records, expire_at_set: expire_at_set, epool_id: epool_id} = state) do
+    if map_size(records) == @max_records_num do
+      {{_expire_at, key}, new_expire_at_set} = :gb_sets.take_largest(expire_at_set)
+      Memcache.delete(key, epool_id)
+      new_records = Map.delete(records, key)
+      %{state | records: new_records, expire_at_set: new_expire_at_set}
+    else
+      state
+    end
+  end
+
+  defp do_write(%{records: records, expire_at_set: expire_at_set, epool_id: epool_id} = state, key, value, lifetime, ratio) do
+    now            = Time.now()
+    expire_at      = Time.shift_seconds(now, lifetime)
+    prob_expire_at = Time.shift_seconds(now, round(lifetime * ratio))
+    Memcache.write(key, value, expire_at, prob_expire_at, epool_id)
+    new_expire_at_set = insert_new_expire_at(expire_at_set, expire_at, key, records)
+    new_records       = Map.put(records, key, expire_at)
+    %{state | records: new_records, expire_at_set: new_expire_at_set}
+  end
+
+  defp insert_new_expire_at(expire_at_set, expire_at, key, records) do
+    new_expire_at_set =
+      case Map.fetch(records, key) do
+        :error               -> expire_at_set
+        {:ok, old_expire_at} -> :gb_sets.delete({old_expire_at, key}, expire_at_set)
+      end
+    :gb_sets.insert({expire_at, key}, new_expire_at_set)
+  end
+
+  #
+  # Public API
+  #
+  defun write(key                 :: term,
+              value               :: term,
+              epool_id            :: v[EPoolId.t],
+              lifetime_in_sec     :: integer,
+              prob_lifetime_ratio :: float) :: :ok | {:error, :too_large_object} do
+    GenServer.call(RegName.memcache_manager(epool_id), {:write, key, value, lifetime_in_sec, prob_lifetime_ratio})
+  end
+end
--- a/core/sup_tree_core/executor_pool/memcache_manager.ex
+++ b/core/sup_tree_core/executor_pool/memcache_manager.ex
@@ -12,7 +12,6 @@ defmodule AntikytheraCore.ExecutorPool.MemcacheManager do
   @max_record_size 65536
 
   use GenServer
-  alias Antikythera.Time
   alias AntikytheraCore.Ets.Memcache
   alias AntikytheraCore.ExecutorPool.RegisteredName, as: RegName
   alias Antikythera.ExecutorPool.Id, as: EPoolId
@@ -28,7 +27,7 @@ defmodule AntikytheraCore.ExecutorPool.MemcacheManager do
 
   @impl true
   def handle_call({:write, key, value, lifetime, ratio}, _from, state) do
-    evicted_state = evict_expired_records(state, Time.now())
+    evicted_state = evict_expired_records(state, System.monotonic_time(:milliseconds))
     if :erts_debug.flat_size(value) > @max_record_size do
       {:reply, {:error, :too_large_object}, evicted_state, @timeout}
     else
@@ -42,7 +41,7 @@ defmodule AntikytheraCore.ExecutorPool.MemcacheManager do
 
   @impl true
   def handle_info(:timeout, state) do
-    new_state = evict_expired_records(state, Time.now())
+    new_state = evict_expired_records(state, System.monotonic_time(:milliseconds))
     {:noreply, new_state, @timeout}
   end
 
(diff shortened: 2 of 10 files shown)


skirino:
CircleCI reported the following test failure:
```
AntikytheraCore.ExecutorPoolTest
  * test should correctly setup processes within the supervision tree structure (0.6ms)

  1) test should correctly setup processes within the supervision tree structure (AntikytheraCore.ExecutorPoolTest)
     test/core/sup_tree_core/executor_pool/executor_pool_test.exs:13
     Assertion with == failed
     code:  assert length(child_pids) == 6
     left:  7
     right: 6
     stacktrace:
       test/core/sup_tree_core/executor_pool/executor_pool_test.exs:27: (test)
```
Just changing the expected value `6` to `7` should fix the test.

Please do run `mix test` on your PC and fix any issues before making a PR.
PR review isn't the right place to address this kind of issues.

lib/util/memcache.ex
```diff
@@ -0,0 +1,60 @@
...
+  defun read(key :: term, epool_id :: v[EPoolId.t]) :: R.t(term, :not_found) do
+    case Memcache.read(key, epool_id) do
+      {:ok, {_, expire_at, prob_expire_at, value}} ->
+        if is_expired(expire_at, prob_expire_at) do
+          {:error, :not_found}
+        else
+          {:ok, value}
+        end
+      {:error, :not_found} -> {:error, :not_found}
+    end
```
skirino: Let's use `R.bind/2`:
```ex
Memcache.read(key, epool_id)
|> R.bind(fn {_, expire_at, prob_expire_at, value} ->
  if is_expired(expire_at, prob_expire_at) do
    {:error, :not_found}
  else
    {:ok, value}
  end
end)
```

lib/util/memcache.ex
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  defun write(key                 :: term,
+              value               :: term,
+              epool_id            :: v[EPoolId.t],
+              lifetime_in_sec     :: v[non_neg_integer] \\ @default_lifetime,
```
skirino: I think the default value for lifetime isn't necessary here; users of this function should be aware of the lifetime of the cache.
Let's make them explicitly specify the lifetime by removing the default.

core/sup_tree_core/executor_pool/memcache_manager.ex
```diff
@@ -0,0 +1,104 @@
...
+      Memcache.delete(key, epool_id)
+      new_records = Map.delete(records, key)
+      %{state | records: new_records, expire_at_set: new_expire_at_set}
+    else
+      state
+    end
+  end
+
+  defp do_write(%{records: records, expire_at_set: expire_at_set, epool_id: epool_id} = state, key, value, lifetime, ratio) do
+    now            = Time.now()
```
skirino: For internal representation of timestamps, isn't it possible to use `pos_integer` obtained from `System.system_time(:seconds)` (or `:milliseconds`)?
It'd make the ETS records smaller.

s417-lama: Indeed it is.
I think `System.monotonic_time/1` is more desirable.

lib/util/memcache.ex
```diff
@@ -0,0 +1,60 @@
...
+              prob_lifetime_ratio :: v[NormalizedFloat.t] \\ @default_ratio) :: :ok | {:error, :too_large_object} do
+    MemcacheManager.write(key, value, epool_id, lifetime_in_sec, prob_lifetime_ratio)
+  end
+
+  defp is_expired(expire_at, prob_expire_at) do
```
skirino: Let's rename this predicate function to `expired?` to make it consistent with the Elixir's naming convention (`is_*` functions are allowed in guards).

skirino: Also this function is called only from `read/2`, so it'd be better to place this function right after `read/2`.

core/sup_tree_core/executor_pool/memcache_manager.ex
skirino: Right, monotonic time is better :+1:

lib/util/memcache.ex
s417-lama: in 282e91f79e0ee0aa32c5c1a5ea680e34b89a665e

lib/util/memcache.ex
s417-lama: in 282e91f79e0ee0aa32c5c1a5ea680e34b89a665e

core/sup_tree_core/executor_pool/memcache_manager.ex
s417-lama: in 282e91f79e0ee0aa32c5c1a5ea680e34b89a665e

lib/util/memcache.ex
s417-lama: in 282e91f79e0ee0aa32c5c1a5ea680e34b89a665e

s417-lama:
Sorry, I forgot to run tests of Antikythera itself.

fixed in b736a79e06a0e2e51cc8a00a03c5ef35de5fe0a7

core/sup_tree_core/executor_pool/memcache_manager.ex
```diff
@@ -0,0 +1,103 @@
+# Copyright(c) 2015-2018 ACCESS CO., LTD. All rights reserved.
+
+use Croma
+
+defmodule AntikytheraCore.ExecutorPool.MemcacheManager do
```
skirino: This module (and the corresponding `GenServer`s) is used only when an object is written as a cache; it's not used for reads.
Should we rename this to `MemcacheWriter` in this regard?

core/sup_tree_core/executor_pool/memcache_manager.ex
```diff
@@ -0,0 +1,103 @@
...
+
+  @impl true
+  def handle_call({:write, key, value, lifetime, ratio}, _from, state) do
+    evicted_state = evict_expired_records(state, System.monotonic_time(:milliseconds))
+    if :erts_debug.flat_size(value) > @max_record_size do
+      {:reply, {:error, :too_large_object}, evicted_state, @timeout}
+    else
+      new_state =
+        evicted_state
+        |> evict_head_record_if_exceeds_limit()
```
skirino: We shouldn't evict the record at head when the key already exists in the cache (in this case number of records don't change).
Let's call `evict_head_record_if_exceeds_limit` from within `do_write/5`.

core/sup_tree_core/executor_pool/memcache_manager.ex
```diff
@@ -0,0 +1,103 @@
...
+
+  @impl true
+  def init(epool_id) do
+    {:ok, %{epool_id: epool_id, records: %{}, expire_at_set: :gb_sets.new()}, @timeout}
+  end
+
+  @impl true
+  def handle_call({:write, key, value, lifetime, ratio}, _from, state) do
+    evicted_state = evict_expired_records(state, System.monotonic_time(:milliseconds))
+    if :erts_debug.flat_size(value) > @max_record_size do
```
skirino: Isn't it possible to off-load this value size checks to caller processes, so that this `GenServer`'s task is minimized?
I think we probably won't have any needs for making "max value size" customizable for each epool.
(If we would then the value size checks should be done in this `GenServer`.)
On the other hand we probably will make "max number of cached objects" customizable, but this is out of scope of this PR.

core/sup_tree_core/executor_pool/memcache_manager.ex
```diff
@@ -0,0 +1,103 @@
...
+        new_records = Map.delete(records, key)
+        evict_expired_records(%{state | records: new_records, expire_at_set: new_expire_at_set}, current_time)
+      else
+        state
+      end
+    end
+  end
+
+  defp evict_head_record_if_exceeds_limit(%{records: records, expire_at_set: expire_at_set, epool_id: epool_id} = state) do
+    if map_size(records) == @max_records_num do
```
skirino: `==` should be `>=` here.
For example when we change `@max_record_num` with hot code upgrade from 10 to 5 and at that time there're 7 cached records, then clients can store as many caches as they want.

core/sup_tree_core/executor_pool/memcache_manager.ex
```diff
@@ -0,0 +1,103 @@
...
+use Croma
+
+defmodule AntikytheraCore.ExecutorPool.MemcacheManager do
+  @moduledoc """
+  A `GenServer` to manage memcache for each executor pool.
+  """
+
+  @timeout         300_000
+  @max_records_num 10
+  @max_record_size 65536
```
skirino: To make it a bit more accurate let's rename `@max_record_size` to `@max_value_size`.

skirino:
Let's declare `key` and `value` as public types.
I think
- we can restrict keys to be not-too-long binaries (strings).
- for values we accept arbitrary term as long as it's size doesn't exceed the limit.

We can encode the restrictions above into the type modules as follows:
```ex
defmodule Key do
  @max_size 100

  @type t :: String.t
  defun valid?(key :: term) :: boolean do
    is_binary(key) and byte_size(key) <= @max_size
  end
end

defmodule Value do
  @max_size 65536

  @type t :: any
  defun valid?(value :: term) :: boolean do
    :erts_debug.flat_size(value) <= @max_size
  end
end
```

With the modules above in hand, we can use `v[Key.t]`, but for `write/4` we should explicitly call `valid?/1`  so that the checks are actually done in cloud environments.

skirino:
Could you add tests for this?
We do need an executor pool, but we basically don't need a concrete gear implementation.
Based on [an existing test case](https://github.com/access-company/antikythera/blob/master/test/core/sup_tree_core/executor_pool/executor_pool_test.exs), try something like the following:
```ex
defmodule Antikythera.MemcacheTest do
  use Croma.TestCase
  alias AntikytheraCore.ExecutorPool
  alias AntikytheraCore.ExecutorPool.Setting, as: EPoolSetting

  @epool_id {:gear, :testgear}

  setup_all do
    ExecutorPool.start_executor_pool(@epool_id, EPoolSetting.default())
    on_exit(fn ->
      ExecutorPoolHelper.kill_and_wait(@epool_id)
    end)
  end

  test "..." do
    ...
  end
end
```

core/sup_tree_core/executor_pool/memcache_manager.ex
s417-lama: How about calling `evict_head_record_if_exceeds_limit` after `do_write`?
It's more concise but exceeds the limit in a moment.

core/sup_tree_core/executor_pool/memcache_manager.ex
s417-lama: Oh, I didn't care about the case.
I think we should call `evict_head_record_if_exceeds_limit` recursively if it is the case.

s417-lama:
Should we restrict the key to string?
I think it would be more useful to accept any terms in the same manner of ETS.

core/sup_tree_core/executor_pool/memcache_manager.ex
skirino: Yeah, changing order of the function calls suffices. Go ahead!

s417-lama:
> With the modules above in hand, we can use v[Key.t], but for write/4 we should explicitly call valid?/1 so that the checks are actually done in cloud environments.

Doesn't `v[...]` always validate the value at runtime?

core/sup_tree_core/executor_pool/memcache_manager.ex
skirino: Right, let's discard all extra records to keep number of records within the limit.

skirino:
> Doesn't v[...] always validate the value at runtime?

Sorry I forgot to mention these:
- https://github.com/access-company/antikythera/blob/master/mix_common.exs#L202
- https://hexdocs.pm/croma/Croma.Defun.html#defun/2-validating-arguments-and-return-value-based-on-their-types

Although it's configurable by antikythera instances, validations such as `x :: v[X.t]` are by default disabled in cloud deployments (treated in the same way as `x :: X.t`),
because its primary purpose is to detect logic errors (programmer's mistakes) about data types during development/test.

skirino:
> Should we restrict the key to string?
I think it would be more useful to accept any terms in the same manner of ETS.

This is fair enough and needs a discussion (sorry I should've explained the intention in the previous comment).
My comment was based on the followings:

- We need to check key size (as well as value size). Based on our current algorithm, each key is stored in the following 3 locations and thus very large key would make trouble.
    1. record in ETS
    2. "set" of GenServer state
    3. "map" of GenServer state

  With `String.t` (i.e. binary) size check is much simpler:
    - taking size of a binary is `O(1)`.
    - with `:erts_debug.flat_size/1` it's difficult to say e.g. whether adding a key-value pair to a map exceeds the limit or not.
- We provide `read` and `write` (i.e. point lookup/update) and don't provide functions such as `list`.
Restricting keys to be `String.t` introduces no change in capability of users; just slight inconvenience of users who already have appropriate but non-`String.t` keys.

Thoughts or comments?

skirino:
> Btw, the size of binary returned by :erts_debug.flat_size/1 seems to be the reference size, not the size of binary itself.

Ah, I misunderstood `:erts_debug.flat_size/1`.
Then checking the inaccurate "size" returned by `:erts_debug.flat_size/1` is pointless.
Thank you for pointing that out! (and sorry for the confusions).

What we can do are
1. Implement a function that computes more accurate memory size of a term
    - `:erlang.term_to_binary() |> byte_size()` or traverse the whole term and aggregate the size
1. Forget about the size check!

I'm now (of course) inclined to (2).
If we don't check value size, key size shouldn't be checked.
Then my slight benefit above of using `String.t` for keys (simpler size checking) disappears.
So let's accept any term for keys.

Concerning the type declarations, we don't need type modules for keys and values, just adding `@type` should be OK.
```ex
@type key :: term
@type value :: term
```

s417-lama:
Or, how about restricting the value to string?
I'm not so clear about what purposes the memcache would be used in, but I guess they want to cache relatively big object (large binary) in memory.

core/sup_tree_core/executor_pool/memcache_manager.ex
s417-lama: Renamed in 88bba1184d0afebfc609c901ffafd5b2f37454ae

core/sup_tree_core/executor_pool/memcache_manager.ex
s417-lama: Fixed in 6ae7fc15a013142f6442c519f27080b81a63bbc9

core/sup_tree_core/executor_pool/memcache_manager.ex
s417-lama: The limit was removed.

core/sup_tree_core/executor_pool/memcache_manager.ex
s417-lama: Fixed in 6ae7fc15a013142f6442c519f27080b81a63bbc9

core/sup_tree_core/executor_pool/memcache_manager.ex
s417-lama: The limit was removed.

s417-lama:
For now, the size limit is removed.
in 6ae7fc15a013142f6442c519f27080b81a63bbc9

s417-lama:
Also, I wrote tests for the memcache.
9035f1d5c23878f1a059bcce67a080a60f0a51c2

core/sup_tree_core/executor_pool/memcache_writer.ex
```diff
@@ -0,0 +1,100 @@
...
+  def handle_info(:timeout, state) do
+    new_state = evict_expired_records(state, System.monotonic_time(:milliseconds))
+    {:noreply, new_state, @timeout}
+  end
+
+  defp evict_expired_records(%{records: records, expire_at_set: expire_at_set, epool_id: epool_id} = state, current_time) do
+    if :gb_sets.is_empty(expire_at_set) do
+      state
+    else
+      {{expire_at, key}, new_expire_at_set} = :gb_sets.take_smallest(expire_at_set)
```
skirino: Oh, my bad.
Very sorry for writing a comment in the ticket about `take_largest`.

s417-lama: No problem. I should notice at first of writing.

skirino:
> Or, how about restricting the value to string?
I'm not so clear about what purposes the memcache would be used in, but I guess they want to cache relatively big object (large binary) in memory.

If we restrict values to be binary then users of memcache who want to cache non-binary terms would end up converting terms to binaries using e.g. `Poison.encode/1` or `:erlang.term_to_binary/1`.
That would introduce extra decoding step when reading a cache.

So... how about checking value size with something like the following function?
Here `size/1` isn't needed for our purpose but it may be used by gear devs to check the size of their values.
Again sorry for the confusion!
```ex
use Croma

defmodule Antikythera.TermUtil do
  defun size(term :: term) :: non_neg_integer do
    :erts_debug.flat_size(term) + total_binary_size(term)
  end

  defunp total_binary_size(term :: term) :: non_neg_integer do
    b when is_bitstring(b)         -> byte_size(b)
    t when is_tuple(t)             -> Tuple.to_list(t) |> Enum.reduce(0, fn(e, acc) -> total_binary_size(e) + acc end)
    l when is_list(l) or is_map(l) -> Enum.reduce(l, 0, fn(e, acc) -> total_binary_size(e) + acc end)
    _                              -> 0
  end

  defun size_smaller_or_equal?(term :: term, limit :: non_neg_integer) :: boolean do
    case limit - :erts_debug.flat_size(term) do
      new_limit when new_limit >= 0 -> limit_minus_total_binary_size(term, new_limit) >= 0
      _                             -> false
    end
  end

  defunp limit_minus_total_binary_size(term :: term, limit :: non_neg_integer) :: integer do
    case term do
      b when is_bitstring(b)         -> limit - byte_size(b)
      t when is_tuple(t)             -> Tuple.to_list(t) |> reduce_while_positive(limit)
      l when is_list(l) or is_map(l) -> reduce_while_positive(l, limit)
      _                              -> limit
    end
  end

  defp reduce_while_positive(enum, limit) do
    Enum.reduce_while(enum, limit, fn(e, l1) ->
      l2 = limit_minus_total_binary_size(e, l1)
      if l2 < 0, do: {:halt, l2}, else: {:cont, l2}
    end)
  end
end
```

s417-lama:
It looks nice to abort immediately after it exceeds the limit!
I'll implement by this method.

Well, should we restrict the key size?

skirino:
> Well, should we restrict the key size?

Hmm... I'm not sure, but let's check also size of keys and protect the system from very rude code that passes a long key.

> The return value of :erts_debug.flat_size/1 seems to be a word size.

Good catch again!
Please include the fixed version of `Antikythera.TermUtil` in this PR.

s417-lama:
Sure.
I implemented the limits for the value and key.
41653b678cb77debea88a455c2af731c18352029

test/lib/util/memcache_test.exs
```diff
@@ -0,0 +1,71 @@
...
+defmodule Antikythera.MemcacheTest do
+  use Croma.TestCase
+  alias AntikytheraCore.ExecutorPool
+  alias AntikytheraCore.ExecutorPool.Setting, as: EPoolSetting
+  alias Antikythera.TermUtil
+
+  @epool_id        {:gear, :testgear}
+  @max_records_num 10
+  @max_key_size    128
+  @max_value_size  65536
```
skirino: Let's add `def` just to retrieve the constants.
```ex
alias Antikythera.Memcache.{Key, Value}
alias AntikytheraCore.ExecutorPool.MemcacheWriter

@max_records_num MemcacheWriter.max_records()
@max_key_size    Key.max_size()
@max_value_size  Value.max_size()
```

core/sup_tree_core/executor_pool/memcache_writer.ex
```diff
@@ -0,0 +1,100 @@
...
+
+use Croma
+
+defmodule AntikytheraCore.ExecutorPool.MemcacheWriter do
+  @moduledoc """
+  A `GenServer` to manage memcache for each executor pool.
+  """
+
+  @timeout         300_000
+  @max_records_num 10
```
skirino: For me `10` seems too few, but I don't have any data how many we should accept...
Can we start with `100` and see what users say about the limit?

test/lib/util/memcache_test.exs
s417-lama: fixed in 60b7e30ab96720c852fce613a29e99e0801de475

core/sup_tree_core/executor_pool/memcache_writer.ex
s417-lama: I didn't consider the number so much.
Ok, I changed the value to 100, in 60b7e30ab96720c852fce613a29e99e0801de475

skirino:
LGTM!
Documentation of the feature can of course be deferred to a separate PR.

ymtszw:
Save https://github.com/access-company/antikythera/pull/14#discussion_r198356487, everything Okay 🚀

core/ets/memcache.ex
```diff
@@ -0,0 +1,33 @@
...
+defmodule AntikytheraCore.Ets.Memcache do
+  alias Croma.Result, as: R
+  alias Antikythera.ExecutorPool.Id, as: EPoolId
+
+  defun init() :: :ok do
+    AntikytheraCore.Ets.create_read_optimized_table(table_name())
+  end
+
+  defun table_name() :: atom do
+    :antikythera_memcache
```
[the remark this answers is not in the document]
s417-lama: Fixed in d120c52902c92c5b89c803d3ac8b9cadd4b44515

PR #15, opened 2018-06-27
[133936] Add docs for Memcache and TermUtil

s417-lama:
https://acsmine.tok.access-company.com/redmine/issues/133936


lib/util/memcache.ex
```diff
@@ -6,10 +6,50 @@ defmodule Antikythera.Memcache do
   @moduledoc """
   In-memory cache for each executor pool.
 
-      iex> Antikythera.Memcache.write("foo", "bar", epool_id)
+  Antikythera provides in-memory cache controlled by AntikytheraCore for each executor pool.
```
ymtszw: I think merging this statement with the above header line would be tidier. 
Also, "controlled by AntikytheraCore" is somewhat "in-depth", and not particularly required here IMO.

How about:

```ex
@moduledoc """
Easy-to-use in-memory cache for each executor pool.

Objects are stored in ETS internally. 
```

(Also, words on "per-instance"-ness could be stated around here, however it may be overkill since "stored in ETS" implies that. I want to hear from folks about this)

lib/util/memcache.ex
```diff
@@ -6,10 +6,50 @@ defmodule Antikythera.Memcache do
...
+  Objects are stored in ETS internally.
+
+  ## Usage
+
+  Memcache behaves as a key-value storage.
```
ymtszw: To me, treating "memcache" as a general noun feels not just right, might be better to use module name explicitly?
In that, putting `#{inspect(__MODULE__)}` is handy since it's rendered as "Antikythera.Memcache" in ex_doc.

```
#{inspect(__MODULE__)} behaves as a key-value storage.
```

Also applies to later parts.

lib/util/memcache.ex
```diff
@@ -6,10 +6,50 @@ defmodule Antikythera.Memcache do
...
       :ok
       iex> Antikythera.Memcache.read("foo", epool_id)
       {:ok, "bar"}
+
+  ## Limitation
+
+  The number of records and the size of keys and values is limited.
+
+  - The maximum number of records for each executor pool is #{AntikytheraCore.ExecutorPool.MemcacheWriter.max_records()}.
+    - If exceeds the limit, records whose expiration time is nearest are evicted.
```
ymtszw: Please use 4-space indents in markdown sub-lists.

(I couldn't just find the canonical spec though. At least we are uniformly using 4-space.)

skirino: I've been advocating the 4-space indentation for markdown lists; you guys should stick with it :)

The reason is simple: layout of bullet/numbered list should reflect its structure even with multi-line items.
- Good (structure of list is visually obvious)
  ```
  - Long text 1 ...
    continued ...
  - Long text 2 ...
    continued ...
  ```
  Bad
  ```
  - Long text 1
  continued ...
  - Long text 2 ...
  continued ...
  ```
- Good
  ```
  - Long text 1 ...
    continued ...
      - Sub item A
      - Sub item B
  ```
  Bad (lines in different levels start at the same position)
  ```
  - Long text 1 ...
    continued ...
    - Sub item A
    - Sub item B
  ```

lib/util/memcache.ex
```diff
@@ -20,6 +60,11 @@ defmodule Antikythera.Memcache do
   alias AntikytheraCore.Ets.Memcache
   alias Antikythera.Memcache.{Key, Value, NormalizedFloat}
 
+  @doc """
+  Read the value associated with the `key` from memcache.
+
+  Please note that records in memcache could be evicted anytime so the error handling must be needed.
+  """
   defun read(key :: term, epool_id :: v[EPoolId.t]) :: R.t(term, :not_found) do
```
skirino: The typespec can be improved here:
- `key :: Key.t`
- return value `:: R.t(Value.t, :not_found)`

lib/util/memcache.ex
```diff
@@ -6,10 +6,50 @@ defmodule Antikythera.Memcache do
   @moduledoc """
   In-memory cache for each executor pool.
 
-      iex> Antikythera.Memcache.write("foo", "bar", epool_id)
+  Antikythera provides in-memory cache controlled by AntikytheraCore for each executor pool.
+  Objects are stored in ETS internally.
```
skirino: We should stick with the same terminology for the cached values.
In most part "record" is used; here it's "object".

Combining @ymtszw's comment above, try this:
```
Easy-to-use in-memory cache for each executor pool.

#{inspect(__MODULE__)} behaves as a key-value storage.
Cached key-value pairs are internally stored in ETS.
It accepts arbitrary (but not too large) terms as both keys and values.
```

lib/util/memcache.ex
s417-lama: Fixed in e8adcb50f657f23724421e7b0383d11d95fe9bb0

I think "in-memory cache" already implies that distributed nodes don't share cached records.

lib/util/memcache.ex
s417-lama: Fixed in e8adcb50f657f23724421e7b0383d11d95fe9bb0

lib/util/memcache.ex
s417-lama: Indeed, it's so reasonable to use 4-space indentation!

Fixed in e8adcb50f657f23724421e7b0383d11d95fe9bb0

lib/util/memcache.ex
s417-lama: I forgot to change them..

Fixed in e8adcb50f657f23724421e7b0383d11d95fe9bb0

lib/util/memcache.ex
s417-lama: FIxed in e8adcb50f657f23724421e7b0383d11d95fe9bb0

lib/util/memcache.ex
```diff
@@ -6,10 +6,50 @@ defmodule Antikythera.Memcache do
...
+  ## Usage
+
+  Memcache behaves as a key-value storage.
+
+      iex> Antikythera.Memcache.write("foo", "bar", epool_id, 3_600)
       :ok
       iex> Antikythera.Memcache.read("foo", epool_id)
       {:ok, "bar"}
+
+  ## Limitation
```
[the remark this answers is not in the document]
s417-lama: Change it to "Limits" in e8adcb50f657f23724421e7b0383d11d95fe9bb0

skirino:
OK!

lib/util/memcache.ex
```diff
@@ -4,12 +4,51 @@ use Croma
...
+  ## Usage
+
+      iex> Antikythera.Memcache.write("foo", "bar", epool_id, 3_600)
       :ok
       iex> Antikythera.Memcache.read("foo", epool_id)
       {:ok, "bar"}
+
+  ## Limits
+
+  The number of records and the size of keys and values is limited.
```
ymtszw: Should be "... are limited."

s417-lama: Fixed in 530f5de234e4c4a344ca0e261f82e7c5ece7e0fb

lib/util/memcache.ex
```diff
@@ -4,12 +4,51 @@ use Croma
...
+
+  - The maximum number of records for each executor pool is #{AntikytheraCore.ExecutorPool.MemcacheWriter.max_records()}.
+      - If exceeds the limit, records whose expiration time is nearest are evicted.
```
ymtszw: This characteristic is wee bit tricky to convey properly, could use a little improvements to my eyes.

Let me clarify: Do you want to state (or imply) "multiple records can be evicted AT ONCE on a single write" here?
I think this is true (although rarely, perhaps on highly concurrent write?), looking at implementation of `evict_head_records_if_exceeds_limit/1`.

If so, how about: "If exceeds the limit, records nearest to expiration are evicted so that a new record can be inserted."
Suggestions welcomed.

s417-lama: Usually, only one record is evicted if exceeds the limit because there is only one writer process (`AntikytheraCore.ExecutorPool.MemcacheWriter`).
However, changing the maximum record number with hot upgrades will cause eviction of multiple records (see this conversation https://github.com/access-company/antikythera/pull/14#discussion_r196703377).

So, eviction of multiple records can rarely occurs and it would be better to change "records" to "a record".

ymtszw: Thanks for redirection, I forgot about https://github.com/access-company/antikythera/pull/14#discussion_r196703377
Yes, if so changing to "a record" should be ok.

lib/util/memcache.ex
```diff
@@ -4,12 +4,51 @@ use Croma
...
+      - If exceeds the limit, `write/5` returns an error `:too_large_key` or `:too_large_value`.
+
+  ## Lifetime of records
+
+  There are 2 cases where records in #{inspect(__MODULE__)} are evicted:
+
+  1. Records are expired (see [Mechanism of Expiration](#module-mechanism-of-expiration) below for more details)
+  2. Reach the maximum number of records for each executor pool
+
+  If it is the case of (2), records whose expiration time is nearest are evicted to keep the maximum number of records.
```
ymtszw: This statement is basically a duplication from the previous section, better be just directing there from a list item?

```
2. Reach the maximum number of records for each executor pool (as mentioned in [Limits](#limits))
```

lib/util/memcache.ex
```diff
@@ -4,12 +4,51 @@ use Croma
...
+
+  Please note that records in #{inspect(__MODULE__)} could be evicted anytime.
+
+  ## Mechanism of Expiration
+
+  The lifetime of records must be set as `lifetime_in_sec` in `write/5`.
+  This lifetime does not guarantee that records remain in the entire specified lifetime.
+
+  To avoid the thundering herd, whether records are expired is decided probabilistically.
```
ymtszw: It would be kind to put a link to some resource about "thundering herd problem" since it is a technical term first mentioned here.
I could not find a very good one but a Wikipedia link should suffice as an entry point: https://en.wikipedia.org/wiki/Thundering_herd_problem

lib/util/memcache.ex
s417-lama: Indeed it is. Fixed in 089fa6487a6319087f67c28dcce9c115e879e845

lib/util/memcache.ex
s417-lama: Added a link in 089fa6487a6319087f67c28dcce9c115e879e845

ymtszw:
Can you edit the explanation figure so that it has white background? :)
Since it gets dark when ex_doc is shown in night mode:

image

s417-lama:
I didn't notice that! I changed the background to white in 5a2b767ac9219de48ea29e200310fb7928396624

ymtszw:
# okay 🚀

PR #17, opened 2018-06-29
[133886] Add processing to retry if Httpc.get returns `:econnrefused error`

ttokit:
https://acsmine.tok.access-company.com/redmine/issues/133886

- Change `fetch_current_version/1` to return error tuple if `Httpc.get/2` returns `:econnrefused`
- Change `wait_until_upgrade_applied_impl/3` to retry on error

### DIFF
--- a/local/running_environment.ex
+++ b/local/running_environment.ex
@@ -3,6 +3,7 @@
 use Croma
 
 defmodule AntikytheraLocal.RunningEnvironment do
+  alias Croma.Result, as: R
   alias Antikythera.{Env, GearNameStr, VersionStr, Httpc}
   alias AntikytheraCore.Path, as: CorePath
   alias AntikytheraLocal.{Cmd, StartScript}
@@ -82,7 +83,8 @@ defmodule AntikytheraLocal.RunningEnvironment do
       raise "Upgrade of #{app_name_str} to #{sha1} hasn't been applied!"
     else
       :timer.sleep(10_000)
-      if fetch_current_version(app_name_str) |> String.ends_with?(sha1) do
+      current_version = fetch_current_version(app_name_str)
+      if R.ok?(current_version) and R.get!(current_version) |> String.ends_with?(sha1) do
         :ok
       else
         wait_until_upgrade_applied_impl(app_name_str, sha1, count + 1)
@@ -93,14 +95,17 @@ defmodule AntikytheraLocal.RunningEnvironment do
   defp fetch_current_version(app_name_str) do
     url = "http://#{AntikytheraCore.Cmd.hostname()}:8080/versions"
     token = File.read!(@system_info_access_token_path)
-    %Httpc.Response{status: 200, body: body} = Httpc.get!(url, %{"authorization" => token})
-    String.split(body, "\n", trim: true)
-    |> Enum.find_value(fn line ->
-      case String.split(line) do
-        [^app_name_str, v] -> v
-        _                  -> nil
-      end
-    end)
+    case Httpc.get(url, %{"authorization" => token}) do
+      {:ok, %Httpc.Response{status: 200, body: body}} ->
+        String.split(body, "\n", trim: true)
+        |> Enum.find_value(fn line ->
+          case String.split(line) do
+            [^app_name_str, v] -> v
+            _                  -> nil
+          end
+        end)
+      {:error, :econnrefused} = error -> error
+    end
   end
 
   defp create_dir_same_as_node() do
--- a/local/running_environment.ex
+++ b/local/running_environment.ex
@@ -100,7 +100,7 @@ defmodule AntikytheraLocal.RunningEnvironment do
         String.split(body, "\n", trim: true)
         |> Enum.find_value(fn line ->
           case String.split(line) do
-            [^app_name_str, v] -> v
+            [^app_name_str, v] -> {:ok, v}
             _                  -> nil
           end
         end)


local/running_environment.ex
```diff
@@ -93,14 +95,17 @@ defmodule AntikytheraLocal.RunningEnvironment do
...
-    |> Enum.find_value(fn line ->
-      case String.split(line) do
-        [^app_name_str, v] -> v
-        _                  -> nil
-      end
-    end)
+    case Httpc.get(url, %{"authorization" => token}) do
+      {:ok, %Httpc.Response{status: 200, body: body}} ->
+        String.split(body, "\n", trim: true)
+        |> Enum.find_value(fn line ->
```
skirino: Use `Enum.find_value/3` instead of `/2` to correctly return `{:error, :not_found}`.
Then we can combine 2 operations (getting HTTP response and extracting version string) by `R.bind/2` as follows:
```ex
Httpc.get(url, %{"authorization" => token})
|> R.bind(fn %Httpc.Response{status: 200, body: body} ->
  String.split(body, "\n", trim: true)
  |> Enum.find_value({:error, :not_found}, fn line ->
    case String.split(line) do
      [^app_name_str, v] -> {:ok, v}
      _                  -> nil
    end
  end)
end)
```
You may (of course) freely make private function(s) to decompose this function.

skirino: Ah, you jut want this code to crash when `app_name_str` doesn't exist in the body.
Then, (although it's about the issue before this PR) let's make the intention explit:

```ex
|> Enum.find(fn line ->
  ...
end) || raise "#{app_name_str} not found in the body!"
```

Here precedence of binary operators is a bit trickey; you may define some helpers here.
(We actually want `Enum.find!`, `Enum.find_value!` etc. here.)

ttokit: fixed in b114f76

Now I have some questions.
- Should we consider the possibility that `line` does not start with `app_name_str`?
For example: `fn line -> String.trim_leading(line) |> String.starts_with?(app_name_str) end`

- Now, I have passed `app_name_str` as an argument for error output only.
However, this seems a subtle way.
Is it OK in this state? Or should we improve?

skirino: First of all, due to my misleading comment (sorry), things got a bit too complicated.
Now we don't need `R.bind/2`; what we want is `R.map/2`.

> Should we consider the possibility that line does not start with app_name_str?

In this case absolutely not!
The response body format is defined by us [here](https://github.com/access-company/antikythera/blob/master/core/handler/system_info_exporter.ex#L37).
Any discrepancies from our expectations should make the tests fail.

local/running_environment.ex
```diff
@@ -93,14 +95,23 @@ defmodule AntikytheraLocal.RunningEnvironment do
...
+
+  defp extract_version_str(body, app_name_str) do
     String.split(body, "\n", trim: true)
-    |> Enum.find_value(fn line ->
-      case String.split(line) do
-        [^app_name_str, v] -> v
-        _                  -> nil
-      end
-    end)
+    |> Enum.find(fn line -> String.starts_with?(line, app_name_str) end)
```
skirino: This could lead to subtle bugs.
Consider the case where `app_name_str` is `"foobar"` and there's a line `"foobarbaz 1.2.3"` in the body.
The problem here is that we must always take delimiters into consideration; only looking at each element is (most of the time) not sufficient.
(And this is one of the most crucial point in string manipulation!)
In this case, `String.starts_with?(line, "#{app_name_str} ")` is OK.

Another minor issue is that we inspect the matched line again at l.110.
It'd be better if we could unite these line manipulations.

So, my recommendation is to define `Antikythera.EnumUtil.find_value!/2`.
(we don't need detailed error message here)
```ex
String.split(body, "\n", trim: true)
|> EnumUtil.find_value!(fn line ->
  case String.split(line) do
    [^app_name_str, v] -> v
    _                  -> nil
  end
end)
```

ttokit: Sorry for misunderstanding your comments.
Implement `Antikythera.EnumUtil.{find!/2, find_value!/2}` in 1509b33,
then I fixed to use it in 3632bda.

Does my implementation match what you expect?

lib/util/enum_util.ex
```diff
@@ -8,7 +8,10 @@ defmodule Antikythera.EnumUtil do
   """
 
   @type context :: any
-  @type item :: any
+  @type item    :: any
+  @type element :: any
```
skirino: Although originally `item` is used to represent each element in an `Enum.t`, `element` looks better.
So let's remove `@type item` here and use `element` throughout this module.
(Actually it's slightly backward-incompatible change to remove a public type, but it's OK to just remove this, since it's almost not used.)

Also this type should have been an alias to `Enum.element`.
```ex
@type element :: Enum.element
```

lib/util/enum_util.ex
```diff
@@ -24,4 +27,24 @@ defmodule Antikythera.EnumUtil do
   defun map_with_context(e :: Enum.t, c :: context, fun :: (item, context -> {item, context})) :: [item] do
     Enum.map_reduce(e, c, fun) |> elem(0)
   end
+
+  @doc """
+  `Enum.find/3` extension for Antikythera.
+
+  If an element is not found, it will crash instead of returning `nil`.
+  Also, only the default value is allowed for the second argument.
```
skirino: Comment on the default value (l.35) is not needed, as the behaviour of this function where no matching element is found is already explained in l.34.
The following 1 line would look OK for me:
```
A variant of `Enum.find/2` that raises an exception when no matching element is found.
```

lib/util/enum_util.ex
ttokit: fixed in 57e0df7

lib/util/enum_util.ex
```diff
@@ -8,7 +8,9 @@ defmodule Antikythera.EnumUtil do
   """
 
   @type context :: any
-  @type item :: any
+  @type element :: Enum.element
+
+  @not_found_error_msg "element not found"
```
skirino: Let's change this error message to:
`"no matching element found"`

local/running_environment.ex
skirino: Yes, now it looks nicer!

lib/util/enum_util.ex
ttokit: fixed in 3f9f446

lib/util/enum_util.ex
ttokit: fixed in 4547392

skirino:
LGTM!

ymtszw:
# okay 🚀

PR #18, opened 2018-06-29
146611_replace_dialyze_with_dialyxir

ymtszw:
https://acsmine.tok.access-company.com/redmine/issues/146611

### DIFF
--- a/doc_src/library_dependencies.md
+++ b/doc_src/library_dependencies.md
@@ -68,10 +68,6 @@
 
 - With [dialyxir](https://github.com/jeremyjh/dialyxir), you can run typecheck of your code by `$ mix dialyzer`.
     - `dialyxir` wraps the `dialyzer`, allowing it to conveniently run from mix projects.
-    - It is recommended that you run with an additional `dialyzer` option:
-        - `--unknown` gives more warnings about unknown functions/types,
-          which usually indicate incorrect type specifications due to e.g. missing aliases.
-          `dialyxir` automatically use this, if not opted-out.
 - See [here](http://learnyousomeerlang.com/dialyzer) for details of success typing analysis.
 
 ### [ex_doc](https://github.com/elixir-lang/ex_doc) for documentations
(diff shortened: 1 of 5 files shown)


doc_src/library_dependencies.md
```diff
@@ -68,7 +68,6 @@
 
 - With [dialyxir](https://github.com/jeremyjh/dialyxir), you can run typecheck of your code by `$ mix dialyzer`.
     - `dialyxir` wraps the `dialyzer`, allowing it to conveniently run from mix projects.
-        - Antikythera itself still uses [dialyze](https://github.com/fishcakez/dialyze), though.
     - It is recommended that you run with an additional `dialyzer` option:
```
skirino: Now we can remove these 4 lines about additional `--unknown` option, can't we?

skirino:
It's almost OK; please address the 2 minor comments and then merge!

doc_src/library_dependencies.md
ymtszw: Mixed feeling about that.

If we consider this part as "assuming you are to use `dialyxir`," then yes, `--unknown` option is used by default in `dialyxir` so we don't have to mention that.

However, this document itself is for a9a instances and gear developers, and admins of a9a instances have freedom not to use `dialyxir` (perhaps using `dialyzer` CLI directly?).
If we are to communicate with such users, it is still beneficial to recommend usage of `--unknown` option IMO.

How about slightly restructuring like this?

```
- With [dialyxir](https://github.com/jeremyjh/dialyxir), you can run typecheck of your code by `$ mix dialyzer`.
    - `dialyxir` wraps the `dialyzer`, allowing it to conveniently run from mix projects.
- If you are using `dialyzer` without `dialyxir`, it is recommended to use `--unknown` option.
    - It gives more warnings about unknown functions/types,
      which usually indicate incorrect type specifications due to e.g. missing aliases.
- See [here](http://learnyousomeerlang.com/dialyzer) for details of success typing analysis.
```

ymtszw:
CONTRIBUTING.md updated 0a22e33

doc_src/library_dependencies.md
skirino: IMO, `dialyxir` has become famous enough (compared with other tools) so that, if an administrator of an antikythera instance determines to go with another tool such as dialyzer CLI, then we can safely assume that he/she already knows about relevant options!

ymtszw: Sounds reasonable. Updated in e285fe4

PR #20, opened 2018-07-05
146116_quit_enforcing_erlang_elixir_versions

ttokit:
https://acsmine.tok.access-company.com/redmine/issues/146116

### DIFF
--- a/doc_src/development_environment.md
+++ b/doc_src/development_environment.md
@@ -7,8 +7,8 @@
 - We strongly recommend [asdf](https://github.com/asdf-vm/asdf) to manage this synchronization.
     - See [antikythera's `.tool-versions` file](https://github.com/access-company/antikythera/blob/master/.tool-versions)
       for the versions of Erlang and Elixir supported in current revision.
-        - Antikythera itself has strict version checking mechanism and raises on compile if different versions are used in your environment.
-        - (This may change later so that Erlang and Elixir versions can be checked by individual antikythera instances, not by antikythera)
+        - Antikythera itself requires Elixir 1.6 or higher, and Erlang's minimum requirement does not exist.
+        - Erlang and Elixir versions can be checked by individual antikythera instances, not by antikythera
 - Other requirements:
     - Some dependencies contain C source code and thus a reasonably new C compiler is required.
     - In macOS you may be warned by outdated GNU make version during compiling antikythera's dependencies.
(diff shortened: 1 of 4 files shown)


skirino:
LGTM!

skirino:
One thing:
Let's explan about this change in `CHANGELOG.md` file, something like:
```
- Next minor version release
    - Strict check of Erlang and Elixir versions is no longer done by antikythera. If an antikythera instance wants to make these versions in sync, it has to check these versions itself.
```

ttokit:
@skirino OK, is minor version releases planned for the future?
(That is, do not write the direct minor version, and leave the description of 0.2.0?)

ttokit:
I fixed `CHANGELOG.md` and document.
Please re-review!

doc_src/development_environment.md
```diff
@@ -7,8 +7,8 @@
 - We strongly recommend [asdf](https://github.com/asdf-vm/asdf) to manage this synchronization.
     - See [antikythera's `.tool-versions` file](https://github.com/access-company/antikythera/blob/master/.tool-versions)
       for the versions of Erlang and Elixir supported in current revision.
-        - Antikythera itself has strict version checking mechanism and raises on compile if different versions are used in your environment.
-        - (This may change later so that Erlang and Elixir versions can be checked by individual antikythera instances, not by antikythera)
+        - Antikythera itself requires Elixir 1.6 or higher, and Erlang's minimum requirement does not exist.
```
ymtszw: I think this line can be deleted.
Two reasons: (1) Elixir version requirement is ever-changing by antikythera versions, so just maintaining the sole canonical instruction - `mix.exs` - would be simpler,
and (2) Erlang version requirement ranges are [provided by Elixir officially](https://hexdocs.pm/elixir/compatibility-and-deprecations.html#compatibility-between-elixir-and-erlang-otp), so not something we must cover.

If there are some specific version combination requirements introduced by our language feature usages then we must note that,
but I cannot think of anything right now. @skirino what do you think?

skirino: Surely we can delete this line.
Actually, now that users don't have to care about the content of `.tool-versions` file, we may delete a bit more.
I think the following should suffice.
```
- We strongly recommend ...
- Antikythera itself doesn't impose any version restrictions other than the one specified in `mix.exs`.
  Individual antikythera instance project may enforce a more fine-grained restrictions on language versions.
```

ttokit: Thank you for the discussion!
fixed in d4d37d4

skirino:
(Again) LGTM!

ymtszw:
# okay 🚀

PR #21, opened 2018-07-12
Move gear_name_test.exs to web/controller/hello_test.exs

SekiT:
Moved the test for hello controller in generated gear, from gear_name_test.exs to web/controller/hello_test.exs.


skirino:
Thanks!

PR #22, opened 2018-07-13
[149372] Handle error due to malformed log messages

ttokit:
https://acsmine.tok.access-company.com/redmine/issues/149372

### DIFF
--- a/core/gear_log/file_handle.ex
+++ b/core/gear_log/file_handle.ex
@@ -21,6 +21,7 @@ defmodule AntikytheraCore.GearLog.FileHandle do
   end
 
   @opaque t :: {Path.t, Time.t, File.io_device}
+  @message_about_malformed_log_message "The write process was skipped because it received malformed data."
 
   defun open(file_path :: Path.t) :: t do
     :ok = File.mkdir_p(Path.dirname(file_path))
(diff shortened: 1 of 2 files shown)


test/core/gear_log/file_handle_test.exs
```diff
@@ -46,6 +46,15 @@ defmodule AntikytheraCore.GearLog.FileHandleTest do
     assert_file_content([log_msg])
   end
 
+  test "write/2 should rescue and continue processing normally" <>
+    "when process crashed by receiving data that is not UTF8 string" do
```
skirino: l.50とl.51の行頭の位置が同じなのは許せません。このレイアウトはやめましょう。
行が長くなっても構わないので単につなげればいいですが、
`"write/2 should skip malformed (non-UTF8) log message"`などと短くしてもOK。

core/gear_log/file_handle.ex
```diff
@@ -48,10 +48,14 @@ defmodule AntikytheraCore.GearLog.FileHandle do
...
-    formatted_lines_str = String.split(msg, "\n", trim: true) |> Enum.map_join(&(prefix <> &1 <> "\n"))
-    IO.write(io_device, formatted_lines_str)
-    write_debug_log(level, formatted_lines_str)
+    try do
+      prefix = log_prefix(time, level, context_id)
+      formatted_lines_str = String.split(msg, "\n", trim: true) |> Enum.map_join(&(prefix <> &1 <> "\n"))
+      IO.write(io_device, formatted_lines_str)
+      write_debug_log(level, formatted_lines_str)
+    rescue
+      _ in ErlangError -> write_debug_log(level, "The write process was skipped because it received malformed data.\n")
```
skirino: - We should match against the reason part `:no_translation`.
In this case the rescued exception is `%ErlangError{original: :no_translation}`, and for this type of exceptions, exception wrapping introduced by `rescue` is (IMO) not very helpful.
We can fall back to the Erlang world by using `catch`.  
- We shouldn't enclose the whole code in `try`. We should enclose only code that may fail.
- When this error occurs we should put a message to the log file, not to tty (console).

So the resulting code should look like:
```ex
try do
  IO.write(io_device, formatted_lines_str)
catch
  :error, :no_translation -> IO.write(io_device, @message_about_malformed_log_message)
end
```

test/core/gear_log/file_handle_test.exs
ttokit: Sorry, it certainly wasn't beautiful.
fixed in fd43d4d

BTW, in case the explanation will unavoidably become long (for example):
```
test "write/2 should rescue and continue processing normally" <>
     "when process crashed by receiving data that is not UTF8 string" do
```
is OK?

skirino: Personally I don't care if a line is a bit too long (code layout and multi-line alignment has precedence), so just put it into one line is OK as long as the test description is reasonable.
The code above is OK for me.
If a long test description really matters then we could also introduce a `describe` block.

ttokit: OK, thank you!

core/gear_log/file_handle.ex
ttokit: fixed in 5045720

- The argument type and return value of `do_write/2` are changed. Because when crashing by receiving an malformed message, the process is already dead. Therefore, it is necessary to open the same log file, then update the handle.
- Also it seems beneficial to me to output error messages not only to the log file but also to the console.

ttokit:
This PR moved to #23

PR #23, opened 2018-07-17
[149372] Remove utf-8 encoding specification

ttokit:
https://acsmine.tok.access-company.com/redmine/issues/149372

### DIFF
--- a/test/core/gear_log/file_handle_test.exs
+++ b/test/core/gear_log/file_handle_test.exs
@@ -46,6 +46,16 @@ defmodule AntikytheraCore.GearLog.FileHandleTest do
     assert_file_content([log_msg])
   end
 
+  test "write/2 should accept malformed (non-UTF8) log message" do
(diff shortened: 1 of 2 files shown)


test/core/gear_log/file_handle_test.exs
```diff
@@ -46,6 +46,16 @@ defmodule AntikytheraCore.GearLog.FileHandleTest do
     assert_file_content([log_msg])
   end
 
+  test "write/2 should accept malformed (non-UTF8) log message" do
+    log_with_malformed_msg = {@log_time, :info, @context_id, <<222>>}
```
skirino: It'd be nicer if we change `<<222>>` to something like `:crypto.strong_rand_bytes(10)`.

skirino:
One minor comment; other than that LGTM

test/core/gear_log/file_handle_test.exs
ttokit: fixed in b26931b

ymtszw:
# okay 🚀

PR #24, opened 2018-07-20
[147879] Add :bypass_job_queue option to AsyncJob to skip addition to job queue

s417-lama:
https://acsmine.tok.access-company.com/redmine/issues/147879


s417-lama:
PR of tests for this change:
https://github.com/access-company/testgear/pull/6

core/async_job/async_job.ex
```diff
@@ -90,18 +92,50 @@ defmodule AntikytheraCore.AsyncJob do
...
     ])
   end
 
+  defunp immediate_job?(options :: v[[option]]) :: boolean do
+    case options[:immediate] do
+      true ->
+        debug_assert(
+          [:schedule, :attempts, :retry_interval]
+          |> Enum.map(fn disallowed -> not Keyword.has_key?(options, disallowed) end)
+          |> Enum.all?(),
```
ymtszw: `Enum.all?/2` can take transformer function so it can be written as:

```ex
Enum.all?([:schedule, :attempts, :retry_interval], fn disallowed -> not Keyword.has_key?(options, disallowed) end)
```

skirino:
Let's first discuss about the name of the option.

The word `:immediate` doesn't look like a good fit for me here because a job enqueued to a distributed job queue can be run "immediately" (if its `:schedule` option says so).
`:immediate` refers to the timing at which the job starts, but actually this option affects not only its scheduling but also other behaviours (e.g. retrying, error due to no available worker).

How do you think about e.g. `:local`?
It's a bit implementation-detail-oriented but (IMO) implies semantic differences.
We can of course use longer option name such as `:execute_in_local_worker`, `:bypass_distributed_job_queue`, etc.

core/async_job/async_job.ex
skirino: - We can use `Enum.all?/2` here.
- Let's return errors by `register/5` when incorrect combinations of options are found, instead of using `debug_assert/2`.

core/async_job/async_job.ex
```diff
@@ -23,6 +24,7 @@ defmodule AntikytheraCore.AsyncJob do
     remaining_attempts: Attempts,
     retry_interval:     RetryInterval,
     payload:            Croma.Map, # opaque data given and used by gear
+    immediate:          Croma.Boolean,
```
skirino: Adding this field to `AntikytheraCore.AsyncJob` struct is rather tricky.
Instances of the current version of this struct (i.e. without `:immediate` field) already exist in the replicated Raft consensus groups and they have to be properly migrated to the new version (i.e. with `:immediate field).
Can we just omit this `:immediate` field?

core/sup_tree_core/executor_pool/async_job_runner.ex
```diff
@@ -142,7 +142,7 @@ defmodule AntikytheraCore.ExecutorPool.AsyncJobRunner do
 
   defp job_succeeded(%{queue_name: queue_name, job_key: job_key} = state) do
     report_on_finish(state, Time.now(), "success")
-    Queue.remove_locked_job(queue_name, job_key)
+    if queue_name, do: Queue.remove_locked_job(queue_name, job_key)
```
skirino: It's very important that now `queue_name` can be `nil` depending on the execution mode.
Let's add comment on l.57 (when filling out `:queue_name` field) as follows:
```ex
queue_name: queue_name, # can be `nil` if the job is executed in `:immediate` mode
```
(Note that, as currently discussed, the name of the option can be changed.)

s417-lama:
Indeed, the name `immediate` is a bit confusing.

I think `bypass_distributed_job_queue` is preferable to `local` or `execute_in_local_worker` that doesn't imply about scheduing, but it is a little long name.
How about `bypass_job_queue` as the option name?

skirino:
Yes, let's go with `:bypass_job_queue`.

core/sup_tree_core/executor_pool/async_job_runner.ex
skirino: In addition to that, we should change typespec of `run/4`:
`queue_name :: v[atom]` => `queue_name :: v[nil | atom]`
(`nil` is an `atom` in a type sense but we should distinguish `nil` from other atoms)

core/async_job/async_job.ex
s417-lama: Thank you, but we ended up to return errors instead of using `debug_assert`.

In 0a17fb1cc0e34322e8281898a5ec061cef1ecb01

s417-lama: Fixed in 0a17fb1cc0e34322e8281898a5ec061cef1ecb01

core/async_job/async_job.ex
s417-lama: Yes, the field is not necessary.

Fixed in 0a17fb1cc0e34322e8281898a5ec061cef1ecb01

core/sup_tree_core/executor_pool/async_job_runner.ex
s417-lama: Fixed in 0a17fb1cc0e34322e8281898a5ec061cef1ecb01

core/async_job/async_job.ex
s417-lama: I wrote tests for it.

a7b43194bf769976a81f4e2ee8c754fa255991d3

ymtszw:
# okay 🚀

skirino:
LGTM!

ttokit:
LGTM!

PR #29, opened 2018-09-10
Add Antikythera.Env.asset_base_url/1

SekiT:
To set `Content-Security-Policy` header in gear application.

I think test should be in testgear.


skirino:
OK, thanks!

PR #30, opened 2018-09-11
[133936] wrapper around foretoken

aMasakiTakahashi:
https://acsmine.tok.access-company.com/redmine/issues/155160


lib/util/foretoken.ex
```diff
@@ -0,0 +1,22 @@
+# Copyright(c) 2015-2018 ACCESS CO., LTD. All rights reserved.
+
+use Croma
+
+defmodule Antikythera.Foretoken do
+  @moduledoc """
+  A thin wrapper around [`Foretoken`](https://github.com/skirino/foretoken) to avoid collisions of bucket name between executor pool.
```
skirino: Let's briefly explain how we avoid collisions:

... avoid collisions between bucket names by prefixing executor pool IDs.

lib/util/foretoken.ex
```diff
@@ -0,0 +1,22 @@
...
+  For details, refer to [Foretoken's documentation](https://hexdocs.pm/foretoken/api-reference.html).
+  """
+
+  @doc """
+  Usage of this function is same as `Foretoken.token/4` except that an executor pool argument is added.
```
skirino: - is same => is the same
- `Foretoken.token/4` => `take/4`
- "an executor pool argument is added" => "an executor pool ID is required as an argument"

mix.exs
```diff
@@ -79,14 +79,14 @@ defmodule Antikythera.Mixfile do
       {:fast_xml    , "1.1.32", [antikythera_internal: true]},
       {:recon       , "2.3.6" , [antikythera_internal: true]},
       {:relx        , "3.23.1", [antikythera_internal: true, only: :prod, runtime: false]}, # only to generate release
+      {:foretoken   , "0.2.0" , [antikythera_internal: true]},
```
skirino: Thanks for this change.
It'd be nicer if `:foretoken` is placed somewhere above `:recon` and `:relx`, as, you know, these 2 hex packages are a bit special.

lib/util/foretoken.ex
```diff
@@ -0,0 +1,22 @@
+# Copyright(c) 2015-2018 ACCESS CO., LTD. All rights reserved.
+
+use Croma
+
+defmodule Antikythera.Foretoken do
```
skirino: Should we name this module as something like `Antikythera.TokenBucket`?
I think we don't have to keep the name of the underlying library here.

lib/util/foretoken.ex
aMasakiTakahashi: Fixed in 1019fb721e18ae6f481e67446d31292a70595834

lib/util/foretoken.ex
aMasakiTakahashi: Fixed in 1019fb721e18ae6f481e67446d31292a70595834

lib/util/foretoken.ex
```diff
@@ -0,0 +1,22 @@
...
+  @doc """
+  Usage of this function is same as `Foretoken.token/4` except that an executor pool argument is added.
+  """
+  defun take(epool_id               :: v[Antikythera.ExecutorPool.Id.t],
+             bucket                 :: any,
+             milliseconds_per_token :: g[pos_integer],
+             max_tokens             :: g[pos_integer],
+             tokens_to_take         :: g[pos_integer] \\ 1) :: :ok | {:error, pos_integer} do
+    bucket_with_epool_id = {epool_id, bucket}
+    Foretoken.Ets.take(bucket_with_epool_id, tokens_to_take, milliseconds_per_token, max_tokens)
```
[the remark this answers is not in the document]
aMasakiTakahashi: Sorry.. `Foretoken.take/4` is correct.
Fixed in fa4ae39da18cf317186689f3e6534560d016cc75

mix.exs
aMasakiTakahashi: Fixed in a4f176fc8b066a78b3351ad1132c3d7759203d9e

lib/util/token_bucket.ex
```diff
@@ -0,0 +1,22 @@
+# Copyright(c) 2015-2018 ACCESS CO., LTD. All rights reserved.
+
+use Croma
+
+defmodule Antikythera.TokenBucket do
+  @moduledoc """
+  A thin wrapper around [`Foretoken`](https://github.com/skirino/foretoken) to avoid collisions between bucket names by prefixing executor pool IDs.
```
skirino: Insert a blank line between l.7 and l.8.

skirino:
Other than https://github.com/access-company/antikythera/pull/30#discussion_r216596156 this looks OK!

ymtszw:
# okay

lib/util/foretoken.ex
aMasakiTakahashi: I think so.
If we use the same name as underlying library, it will be difficult to replace `Foretoken`.
So, I renamed.
Fixed in 07a9f4a74500d9fac41677ee52637e6db7f6965f

lib/util/token_bucket.ex
aMasakiTakahashi: Fixed in 87094bdf3338e52dd4fe86c99edf695f99b3b830

test/lib/util/token_bucket_test.exs
```diff
@@ -0,0 +1,17 @@
+# Copyright(c) 2015-2018 ACCESS CO., LTD. All rights reserved.
+
+defmodule Antikythera.TokenBucketTest do
+  use Croma.TestCase
+  alias Croma.Result, as: R
+  alias Antikythera.TokenBucket
```
ttokit: Since you are using `Croma.TestCase`, you can delete this line.

aMasakiTakahashi: Thank you.
Fixed in 593f228af9bed136734af602491a76dc06496301

PR #31, opened 2018-09-14
[133835] treat payload of async jobs as binary internally

s417-lama:
- temporarily accept non-binary payload on async_job_runner due to hot upgrading

https://acsmine.tok.access-company.com/redmine/issues/133835

### DIFF
--- a/core/async_job/async_job.ex
+++ b/core/async_job/async_job.ex
@@ -39,7 +39,7 @@ defmodule AntikytheraCore.AsyncJob do
       job_id                        <- extract_job_id(options)
       bypass?                       <- bypass_job_queue?(options)
       {schedule, start_time_millis} <- validate_schedule(now_millis, options)
-      job                           <- make_job(gear_name, module, payload, schedule, options)
+      job                           <- make_job(gear_name, module, payload, schedule, bypass?, options)
       do_register(epool_id, job_id, job, bypass?, start_time_millis, now_millis)
     end
   end
@@ -98,8 +98,10 @@ defmodule AntikytheraCore.AsyncJob do
                    module    :: v[module],
                    payload   :: v[map],
                    schedule  :: v[Schedule.t],
+                   bypass?   :: v[boolean],
                    options   :: v[[option]]) :: R.t(t) do
-    attempts = Keyword.get(options, :attempts, Attempts.default())
+    attempts        = Keyword.get(options, :attempts, Attempts.default())
+    encoded_payload = if bypass?, do: payload, else: :erlang.term_to_binary(payload, [:compressed])
     new([
       gear_name:          gear_name,
       module:             module,
@@ -108,7 +110,7 @@ defmodule AntikytheraCore.AsyncJob do
       attempts:           attempts,
       remaining_attempts: attempts,
       retry_interval:     Keyword.get(options, :retry_interval, RetryInterval.default()),
-      payload:            :erlang.term_to_binary(payload, [:compressed]),
+      payload:            encoded_payload,
     ])
   end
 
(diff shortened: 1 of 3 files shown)


ymtszw:
# okay 🚀

ttokit:
LGTM!

core/async_job/async_job.ex
```diff
@@ -108,7 +108,7 @@ defmodule AntikytheraCore.AsyncJob do
       attempts:           attempts,
       remaining_attempts: attempts,
       retry_interval:     Keyword.get(options, :retry_interval, RetryInterval.default()),
-      payload:            payload,
+      payload:            :erlang.term_to_binary(payload),
```
skirino: Compressing payload at the same time could be beneficial in terms of memory space occupied by the replicated job queue.
Let's add `[:compressed]` as the 2nd argument.
http://erlang.org/doc/man/erlang.html#term_to_binary-2

core/sup_tree_core/executor_pool/async_job_runner.ex
```diff
@@ -201,13 +203,16 @@ defmodule AntikytheraCore.ExecutorPool.AsyncJobRunner do
...
+
   #
   # Public functions to run within a separate process
   #
-  defun do_run(module :: v[module], payload :: v[map], metadata :: v[Metadata.t], context :: v[Context.t]) :: any do
+  defun do_run(module :: v[module], payload :: v[map | binary], metadata :: v[Metadata.t], context :: v[Context.t]) :: any do
     ContextHelper.set(context)
     try do
-      module.run(payload, metadata, context)
+      module.run(decode_payload(payload), metadata, context)
```
skirino: Can we decompress the payload only once per job execution?
For example we could simply add a local variable in `handle_cast/2` to hold the decompressed payload so that decoding steps in `log_prefix/3` and `do_run/4` are combined.

Concerning `do_abandon/4`, it's OK to decode on-demand since:
- it should be called on a very rare occasions,
- when called, it can be far apart from `do_run/4` in terms of time (up to 30min); it seems better not to keep the decompressed payload in the heap memory of `AsyncJobRunner`.

skirino:
Thank you for tackling this!

Though the patch looks good, extra care must be taken when deploying this to a working cluster, as upgrades in multiple nodes are not an atomic event.
The patch consists of the following 2 parts:
1. receiver side (`core/sup_tree_core/executor_pool/async_job_runner.ex`): `AsyncJobRunner` starts to recognize the binary-encoded payloads.
2. sender side (`core/async_job/async_job.ex`): New jobs are enqueued in binary format.

And we should deploy (1) separately before (2).

After completion of the review rounds, please merge this PR as 2 separate commits for (1) and (2).

core/async_job/async_job.ex
s417-lama: added in 16e243cf7f137a3adf860871e90a12f8a6454133

core/sup_tree_core/executor_pool/async_job_runner.ex
s417-lama: fixed in 2da715895c07c68e7e3abe20419d994ca535231f

skirino:
LGTM!

s417-lama:
Compressing is not necessary and causes additional overhead when :bypass_job_queue option is specified, so I fixed it.

553ddec0b555eeeefda21aa2cdb4a4b6606c80de

Currently, we accept map and binary as payload in `async_job_runner.ex` due to hot upgrades, but by this fixation, it becomes not required to accept only binary payload after all registered job's payload become binary.

PR #34, opened 2018-10-05
[156276] Set deps dir as an environmental variable in gears' mix.exs

s417-lama:
https://acsmine.tok.access-company.com/redmine/issues/156276

### DIFF
--- a/priv/gear_template/mix.exs
+++ b/priv/gear_template/mix.exs
@@ -1,11 +1,13 @@
 instance_dep = <%= instance_dep %>
 
 try do
-  parent_dir = Path.expand("..", __DIR__)
   deps_dir =
-    case Path.basename(parent_dir) do
-      "deps" -> parent_dir                 # this gear project is used by another gear as a gear dependency
-      _      -> Path.join(__DIR__, "deps") # this gear project is the toplevel mix project
+    case Application.get_env(:antikythera, :deps_dir) do
+      nil ->
+        deps_path = Mix.Project.deps_path()
+        Application.put_env(:antikythera, :deps_dir, deps_path)
+        deps_path
+      deps_path -> deps_path
     end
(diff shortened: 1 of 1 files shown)


priv/gear_template/mix.exs
```diff
@@ -1,11 +1,13 @@
 instance_dep = <%= instance_dep %>
 
 try do
-  parent_dir = Path.expand("..", __DIR__)
   deps_dir =
-    case Path.basename(parent_dir) do
-      "deps" -> parent_dir                 # this gear project is used by another gear as a gear dependency
-      _      -> Path.join(__DIR__, "deps") # this gear project is the toplevel mix project
+    case Application.get_env(:antikythera, :deps_dir) do
```
skirino: This application config is used only during compilation and thus very special.
How about naming this as something like `:compile_time_deps_dir`?
Also I'd need comments to persist our understandings about the issue.
```ex
   deps_dir =
     # Gears use mix settings of both antikythera and antikythera instance and thus paths to these projects must be available.
     # We have to remember `deps_path()` so that gear dependency with `:path` option can reach these projects.
     case Application.get_env(:antikythera, :compile_time_deps_dir) do
       nil -> # this gear project is the toplevel mix project
         deps_path = Mix.Project.deps_path()
         Application.put_env(:antikythera, :compile_time_deps_dir, deps_path)
         deps_path
       deps_path -> deps_path # this gear project is used by another gear as a gear dependency
     end
```

s417-lama: OK, fixed in d1c581f61ea475e45e41c82c8ca79e2e92c84460

skirino:
OK!

ttokit:
LGTM!

s417-lama:
@ymtszw I'm going to merge this PR right now. Please review this when you have enough time.

PR #35, opened 2018-10-15
[157910] Cap `ws_max_connections` in gear/tenant exec pool settings based on available system memory

skirino:
https://acsmine.tok.access-company.com/redmine/issues/157910


ttokit:
LGTM!

PR #40, opened 2018-12-18
[164572] Add a function that receives a fun to be called if there's no memcache

s417-lama:
https://acsmine.tok.access-company.com/redmine/issues/164572

### DIFF
--- a/lib/util/memcache.ex
+++ b/lib/util/memcache.ex
@@ -73,6 +73,31 @@ defmodule Antikythera.Memcache do
     end)
   end
 
+  @doc """
+  Try to read a value associated with the `key` from #{inspect(__MODULE__)}.
+
+  `fun` is evaluated if a value is not found, and the new value returned by `fun` is stored in #{inspect(__MODULE__)}.
+  If a value is found in #{inspect(__MODULE__)} or writing the new value to #{inspect(__MODULE__)} succeeds, the value is returned as `{:ok, value}`, but if writing the new value fails, an error is returned in the same manner as `write/5`.
+
+  Parameters `lifetime_in_sec` and `prob_lifetime_ratio` are used to call `write/5` and the details are described above.
+  """
+  defun read(key                 :: Key.t,
+             epool_id            :: v[EPoolId.t],
+             lifetime_in_sec     :: v[non_neg_integer],
+             fun                 :: (-> Value.t),
+             prob_lifetime_ratio :: v[NormalizedFloat.t] \\ @default_ratio) :: R.t(Value.t, :too_large_key | :too_large_value) do
+    read(key, epool_id)
+    |> case do
+      {:ok, value}         -> {:ok, value}
+      {:error, _not_found} ->
+        value = fun.()
+        case write(key, value, epool_id, lifetime_in_sec, prob_lifetime_ratio) do
+          :ok -> {:ok, value}
+          err -> err
+        end
+    end
+  end
+
   defp expired?(expire_at, prob_expire_at) do
     case System.monotonic_time(:millisecond) do
       now when now < prob_expire_at -> false
--- a/test/lib/util/memcache_test.exs
+++ b/test/lib/util/memcache_test.exs
@@ -38,6 +38,12 @@ defmodule Antikythera.MemcacheTest do
     assert Memcache.read( "foo", @epool_id)                    == {:ok, "hoge"}
   end
 
+  test "should write values to memcache and read them with read/5" do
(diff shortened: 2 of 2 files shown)


lib/util/memcache.ex
```diff
@@ -73,6 +73,31 @@ defmodule Antikythera.Memcache do
...
+  `fun` is evaluated if a value is not found, and the new value returned by `fun` is stored in #{inspect(__MODULE__)}.
+  If a value is found in #{inspect(__MODULE__)} or writing the new value to #{inspect(__MODULE__)} succeeds, the value is returned as `{:ok, value}`, but if writing the new value fails, an error is returned in the same manner as `write/5`.
+
+  Parameters `lifetime_in_sec` and `prob_lifetime_ratio` are used to call `write/5` and the details are described above.
+  """
+  defun read(key                 :: Key.t,
+             epool_id            :: v[EPoolId.t],
+             lifetime_in_sec     :: v[non_neg_integer],
+             fun                 :: (-> Value.t),
+             prob_lifetime_ratio :: v[NormalizedFloat.t] \\ @default_ratio) :: R.t(Value.t, :too_large_key | :too_large_value) do
```
skirino: I think it'd be nicer to flip `fun` and `prob_lifetime_ratio` so that on calling side possibly long `fn` expression comes last in the parameter list.

s417-lama: I have misunderstood that middle args can have default values!
Fixed in 330a2b6cefc2d391f3cf4e0b997e2093bc09b0ca

lib/util/memcache.ex
```diff
@@ -73,6 +73,31 @@ defmodule Antikythera.Memcache do
...
 
+  @doc """
+  Try to read a value associated with the `key` from #{inspect(__MODULE__)}.
+
+  `fun` is evaluated if a value is not found, and the new value returned by `fun` is stored in #{inspect(__MODULE__)}.
+  If a value is found in #{inspect(__MODULE__)} or writing the new value to #{inspect(__MODULE__)} succeeds, the value is returned as `{:ok, value}`, but if writing the new value fails, an error is returned in the same manner as `write/5`.
+
+  Parameters `lifetime_in_sec` and `prob_lifetime_ratio` are used to call `write/5` and the details are described above.
+  """
+  defun read(key                 :: Key.t,
```
skirino: Concerning the function name:
I'd like to somehow express that this function is side-effecting one.
`read_or_write` looks as if "read" and "write" are on an equal footing (which they are not).
How about something like `read_or_else_write_with_fun` (a bit lengthy but hopefully easier to read)?
Any thoughts?

skirino:
(Just dumping my thought on the interface)
There might be cases where `lifetime` and/or `prob_lifetime_ratio` of each cached item is not known when calling this newly-added function; instead it's determined from the info fetched from origin.
Here we are providing a convenience function that simplifies common use cases and thus we ignore such situation.

lib/util/memcache.ex
s417-lama: Indeed, the name is confusing.
`read_or_else_write_with_fun` is well-understandable, but I think `read_or_else_write` is sufficient to express the meaning.

skirino: OK, let's go with `read_or_else_write`!

s417-lama: Renamed in 2e944a195874f417aa0cb8845c1a6a39582f894f

lib/util/memcache.ex
```diff
@@ -73,6 +73,31 @@ defmodule Antikythera.Memcache do
...
+
+  Parameters `lifetime_in_sec` and `prob_lifetime_ratio` are used to call `write/5` and the details are described above.
+  """
+  defun read_or_else_write(key                 :: Key.t,
+                           epool_id            :: v[EPoolId.t],
+                           lifetime_in_sec     :: v[non_neg_integer],
+                           prob_lifetime_ratio :: v[NormalizedFloat.t] \\ @default_ratio,
+                           fun                 :: (-> Value.t)) :: R.t(Value.t, :too_large_key | :too_large_value) do
+    read(key, epool_id)
+    |> case do
```
skirino: Minor issue:
Let's collapse the 2 lines into 1`|>` as 1st arg to `case` macro is sufficiently simple.
```ex
case read(key, epool_id) do
```

skirino:
Please address the 1 minor issue; except for that LGTM!

lib/util/memcache.ex
s417-lama: Fixed in f65471aad7c25f1284de55d7d341aa7de52c525f

PR #41, opened 2019-02-18
[169382] Fix app acquisition at version check

ttokit:
https://acsmine.tok.access-company.com/redmine/issues/169382


skirino:
LGTM

PR #43, opened 2019-04-22
[174208] Fix Antikythera.Url to reject unencoded line break

aNobuhisaMasaki:
https://acsmine.tok.access-company.com/redmine/issues/174208

Fixed URL pattern to reject unencoded line breaks.

Several parts in `Antikythera.Url` reject unencoded `" "`, but not for unencoded `\n`.
In this fix, `" "` in regex pattern is replaced with `\s` to reject unencoded line break.

TODO:
* Think of line breaks in `Antikythera.UnencodedPath`


skirino:
LGTM, thanks!

PR #47, opened 2019-06-03
[177185] Add macro to separate cmd fun

aNobuhisaMasaki:
https://acsmine.tok.access-company.com/redmine/issues/177185

I want to suggest a simple, but questionable fix in `Antikythera.Zip`.

I noticed that `Antikythera.Zip` module ends in experiencing whole turnaround using macro, passing through `&System.cmd("zip", &1)` as the sole parameter.
This is for testing; mainly because `System.cmd/2` cannot be mocked.
Or, are there any other methods for testing?

### DIFF
--- a/lib/util/zip.ex
+++ b/lib/util/zip.ex
@@ -105,19 +105,19 @@ quote do
         end
       end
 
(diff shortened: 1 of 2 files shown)


aNobuhisaMasaki:
I will try another fix.
Maybe it's easier to just pass cd to System.cmd/3.

PR #48, opened 2019-06-04
[177185] Use relative path for zip library

aNobuhisaMasaki:
https://acsmine.tok.access-company.com/redmine/issues/177185

Enabled using relative path by passing `cwd` to `Zip.zip` (now arity 5),
then turned into `:cd` option in `System.cmd/3`.
Otherwise, when absolute paths are used, you will have a zip file *including directory structure*.

I appreciate your suggestions.

### DIFF
--- a/lib/util/zip.ex
+++ b/lib/util/zip.ex
@@ -53,6 +53,7 @@ defmodule Antikythera.Zip do
          :ok           <- validate_within_tmpdir(cwd_path, tmpdir),
          :ok           <- validate_within_tmpdir(zip_path, tmpdir),
          :ok           <- validate_within_tmpdir(src_path, tmpdir),
+         :ok           <- reject_existing_file(cwd_path),
          :ok           <- reject_existing_dir(zip_path),
          :ok           <- ensure_dir_exists(zip_path, tmpdir),
          :ok           <- validate_path_exists(src_path),
@@ -69,6 +70,14 @@ defmodule Antikythera.Zip do
 
   defp extract_epool_id(%Context{executor_pool_id: epool_id}), do: epool_id
   defp extract_epool_id(epool_id),                             do: epool_id
+
+  defunp reject_existing_file(path :: v[String.t]) :: :ok | {:error, tuple} do
+    if File.dir?(path) do
+      :ok
+    else
+      {:error, {:not_dir, %{path: path}}}
+    end
+  end
 
   defunp reject_existing_dir(path :: v[String.t]) :: :ok | {:error, tuple} do
     if File.dir?(path) do
--- a/lib/util/zip.ex
+++ b/lib/util/zip.ex
@@ -88,7 +88,7 @@ defmodule Antikythera.Zip do
   end
 
   defunp validate_within_tmpdir(path :: v[String.t], tmpdir :: v[String.t]) :: :ok | {:error, tuple} do
-    if String.starts_with?(path, "#{tmpdir}/") do
+    if path == tmpdir or String.starts_with?(path, "#{tmpdir}/") do
       :ok
     else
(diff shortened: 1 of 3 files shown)


aNobuhisaMasaki:
I have a trouble in commit order...
For correct ones: https://github.com/access-company/antikythera/pull/48/commits

lib/util/zip.ex
```diff
@@ -41,21 +51,25 @@ defmodule Antikythera.Zip do
     end)
   """
   defun zip(context_or_epool_id :: v[EPoolId.t | Context.t],
+            cwd_raw_path        :: v[String.t],
             zip_raw_path        :: v[FileName.t],
-            src_raw_path        :: v[String.t],
+            src_raw_path        :: v[NonTraversalPath.t],
```
aNobuhisaMasaki: Prohibited non-traversal path for `src_path`, in order to prevent from creating a *fish print* of directories outside `tmpdir`.

aNobuhisaMasaki: > Prohibited non-traversal path for src_path, in order to prevent from creating a fish print of directories outside tmpdir.

In this sense, absolute paths like `/tmp...` are also problematic...

aNobuhisaMasaki: >> Prohibited non-traversal path for src_path, in order to prevent from creating a fish print of directories outside tmpdir.

> In this sense, absolute paths like /tmp... are also problematic...

- Fixed:
  - Invalidated absolute path for src e9a3716
  - Remove auth for src path ddf6018

aNobuhisaMasaki: > Invalidated absolute path for src

Note that `/` cannot be a `tmpdir` (though rarely zipped, anyway).

lib/util/zip.ex
```diff
@@ -113,6 +118,24 @@ defmodule Antikythera.Zip do
...
+    if File.exists?(path) do
+      :ok
+    else
+      {:error, {:not_found, %{path: path}}}
+    end
+  end
+
+  defunp validate_suffix(src_raw_path :: v[String.t], cwd_path :: v[String.t]) :: :ok | {:error, tuple} do
+    suffixed = String.ends_with?(src_raw_path, "/")
+    path = Path.expand(src_raw_path, cwd_path)
```
aNobuhisaMasaki: There are two formats of relationship between names,
- raw_path -> path
- path -> full_path

which are the same in meaning.
I will unify expression to the latter.

aNobuhisaMasaki: Fixed naming rule in c78679f.

aNobuhisaMasaki:
@aMasakiTakahashi @ttokit @toyoaki-k 
Fixed all my concerns; please check out for it.

aNobuhisaMasaki:
@aMasakiTakahashi @ttokit @toyoaki-k
For manual test:
```
$ iex -S mix

iex > 

context = Antikythera.Test.ConnHelper.make_conn().context
Antikythera.Tmpdir.make(context, fn tmpdir ->
  src = tmpdir <> "/t.txt"
  File.write!(src, "text")
  Antikythera.Zip.zip(context, tmpdir, "z.zip", "t.txt")
end)

{:ok, "/YOUR/TMP/DIR/z.zip"}
```
error cases:
```
  Antikythera.Zip.zip(context, tmpdir, "z.zip", "t..txt")
  Antikythera.Zip.zip(context, tmpdir, "z.zip", "/t.txt")
  Antikythera.Zip.zip(context, src, "z.zip", "t.txt")
```

aNobuhisaMasaki:
Closing,
because it is still unclear that it is safe to use the concept of working directory,
especially whilst interaction between Erlang processes.

PR #51, opened 2019-06-24
[179864] fix credo error: large abc size

irisTa56:
ticket: https://acsmine.tok.access-company.com/redmine/issues/179864

This commit passes `mix credo --strict` for the below three files:

* `lib/mix/gear_static_analysis.ex`
* `lib/web/template/precompiler.ex`
* `core/async_job/queue.ex`

Most of the changes are deconstruction of `cond` and `case` in a function by defining same-name functions distinguished by pattern-matching (like overloading); These changes are for reducing ABC size of functions.

PR #56, opened 2019-07-22
179864 credo large abc size revised

irisTa56:
ticket: https://acsmine.tok.access-company.com/redmine/issues/179864

This PR is revised version of #51.
Only one change from #51 is mentioned in https://github.com/access-company/antikythera/commit/2328b92c87a1542a71e9906119340238a9038662#r34391893.

Using `Module.safe_concat/1` here throws `ArgumentError` when a resulting module name includes alias(es) and is used in only type specification.
I guess that atoms for such module names won’t be created at compile time because they won’t be used; Elixir requires such module names for only static analysis.

I think that we can safely skip a check for a module throwing `ArgumentError` because the module is actually used elsewhere, and Antikythera can detect problems there.

### DIFF
--- a/core/async_job/queue.ex
+++ b/core/async_job/queue.ex
@@ -54,20 +54,32 @@ defmodule AntikytheraCore.AsyncJob.Queue do
   @impl true
   defun new() :: t do
     set = :gb_sets.empty()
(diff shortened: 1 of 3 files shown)


ttokit:
LGTM!

aMasakiTakahashi:
LGTM!

PR #102, opened 2019-10-28
[189201] Enable embedded mode

aDaichiOta:
https://acsmine.tok.access-company.com/redmine/issues/189201

### DIFF
--- a/core/version/gear.ex
+++ b/core/version/gear.ex
@@ -31,7 +31,7 @@ defmodule AntikytheraCore.Version.Gear do
         Enum.each(Application.spec(gear_name, :modules), fn(mod) ->
           # Croma defines modules (e.g. `Elixir.Croma.TypeGen.Nilable.Antikythera.Email`) automatically in the gear.
           # As a result, some modules which have the same name are defined in two or more gears.
-          # We have to avoid loading these modules twice.
+          # We should avoid loading these modules twice.
           if !auto_generated_module?(mod) do
             {:module, _} = :code.load_file(mod)
           else
--- a/core/version/gear.ex
+++ b/core/version/gear.ex
@@ -29,17 +29,19 @@ defmodule AntikytheraCore.Version.Gear do
       if :code.get_mode() != :interactive do
         # Load module manually
         Enum.each(Application.spec(gear_name, :modules), fn(mod) ->
-          case :code.load_file(mod) do
-            {:module, _}      -> :ok
-            {:error , reason} ->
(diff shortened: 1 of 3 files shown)


core/version/gear.ex
```diff
@@ -29,15 +29,17 @@ defmodule AntikytheraCore.Version.Gear do
...
-          if not skip_to_load? do
-            case :code.load_file(mod) do
-              {:module, _}      -> :ok
-              {:error , reason} -> raise "Failed to load '#{mod}': #{reason}"
-            end
+          case :code.load_file(mod) do
+            {:module, _}      -> :ok
+            {:error , reason} ->
+              # Croma defines modules (e.g. `Elixir.Croma.TypeGen.Nilable.Antikythera.Email`) automatically in the gear.
+              # As a result, some modules which have the same name are defined in two or more gears, and loading these modules causes an error.
```
irisTa56: (optional)

```suggestion
              # As a result, some modules with the same name are defined in two or more gears, and loading these modules causes an error.
```

I think this is easier to read.

core/version/gear.ex
```diff
@@ -32,12 +32,18 @@ defmodule AntikytheraCore.Version.Gear do
...
+                {:module, _}           -> :ok
+                # When three or more same-name modules are loaded at the same time,
+                # `:code.load_file(mod)` returns `:not_purged`. We can safely ignore the error in that case.
+                {:error , :not_purged} -> :ok
+                {:error , reason}      -> raise "Failed to load '#{mod}': #{reason}"
+              end
             end
+          else
+            {:module, _} = :code.load_file(mod)
           end
```
irisTa56: (optional)
I think writing shorter case first is more readable.

```elixir
          if not auto_generated_module?(mod) do
            {:module, _} = :code.load_file(mod)
          else
            if !:erlang.module_loaded(mod) do
              case :code.load_file(mod) do
                {:module, _}           -> :ok
                # When three or more same-name modules are loaded at the same time,
                # `:code.load_file(mod)` returns `:not_purged`. We can safely ignore the error in that case.
                {:error , :not_purged} -> :ok
                {:error , reason}      -> raise "Failed to load '#{mod}': #{reason}"
              end
            end
          end
```

irisTa56:
Just a comment.
LGTM!

core/version/gear.ex
aDaichiOta: Fixed be13b4df2c711637193077408f3060fea4c9efc4

ttokit:
LGTM!

aMasakiTakahashi:
LGTM!

PR #126, opened 2020-05-19
[202670] Mix task can set the ContextId in GearLog

aYukiSekiguchi:
https://acsmine.tok.access-company.com/redmine/issues/202670

### DIFF
--- a/core/web/context.ex
+++ b/core/web/context.ex
@@ -24,9 +24,13 @@ defmodule AntikytheraCore.Context do
   end
 
   defun make_context_id(t :: v[Time.t]) :: String.t do
+    make_context_id(t, NodeId.get())
+  end
+
+  defun make_context_id(t :: v[Time.t], node_id :: String.t) :: String.t do
     [
       timestamp(t),
-      NodeId.get(),
+      node_id,
       :erlang.pid_to_list(self()) |> tl() |> List.to_string() |> String.trim_trailing(">"),
     ] |> Enum.join("_")
   end
--- a/core/web/context.ex
+++ b/core/web/context.ex
@@ -3,11 +3,11 @@
 use Croma
 
 defmodule AntikytheraCore.Context do
-  alias Antikythera.{Time, GearName, ContextId}
+  alias Antikythera.{Time, GearName, ContextId, NodeId}
   alias Antikythera.Context, as: LContext
   alias Antikythera.Context.GearEntryPoint
   alias Antikythera.ExecutorPool.Id, as: EPoolId
-  alias AntikytheraCore.Cluster.NodeId
+  alias AntikytheraCore.Cluster.NodeId, as: CoreNodeId
 
   defun make(gear_name   :: v[GearName.t],
              entry_point :: v[nil | GearEntryPoint.t],
@@ -24,10 +24,10 @@ defmodule AntikytheraCore.Context do
   end
 
   defun make_context_id(t :: v[Time.t]) :: ContextId.t do
-    make_context_id(t, NodeId.get())
+    make_context_id(t, CoreNodeId.get())
   end
 
-  defun make_context_id(t :: v[Time.t], node_id :: v[String.t]) :: v[ContextId.t] do
+  defun make_context_id(t :: v[Time.t], node_id :: v[NodeId.t]) :: v[ContextId.t] do
     [
       timestamp(t),
       node_id,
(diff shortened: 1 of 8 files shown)


core/web/context.ex
```diff
@@ -24,9 +24,13 @@ defmodule AntikytheraCore.Context do
   end
 
   defun make_context_id(t :: v[Time.t]) :: String.t do
+    make_context_id(t, NodeId.get())
+  end
+
+  defun make_context_id(t :: v[Time.t], node_id :: String.t) :: String.t do
```
irisTa56, line 30:
```suggestion
  defun make_context_id(t :: v[Time.t], node_id :: v[String.t]) :: String.t do
```

lib/mix/task.ex
```diff
@@ -20,4 +24,18 @@ defmodule Antikythera.Mix.Task do
...
+
+  @doc """
+  Set the specified `node_id` to the `Antikythera.ContextId` in GearLog.
+
+  If you want to use GearLog in mix task, you must set `node_id` before you call GearLog functions.
+
+  The `Antikythera.ContextId` in GearLog will become `{timestamp}_{node_id}_{PID}`.
+  The `timestamp` and `PID` are automatically got from system.
+  """
+  defun set_node_id_to_gear_log_context(node_id :: String.t) :: :ok do
```
irisTa56, line 36:
```suggestion
  defun set_node_id_to_gear_log_context(node_id :: v[String.t]) :: :ok do
```

lib/mix/task.ex
```diff
@@ -20,4 +24,18 @@ defmodule Antikythera.Mix.Task do
...
+  Set the specified `node_id` to the `Antikythera.ContextId` in GearLog.
+
+  If you want to use GearLog in mix task, you must set `node_id` before you call GearLog functions.
+
+  The `Antikythera.ContextId` in GearLog will become `{timestamp}_{node_id}_{PID}`.
+  The `timestamp` and `PID` are automatically got from system.
+  """
+  defun set_node_id_to_gear_log_context(node_id :: v[String.t]) :: :ok do
+    context_id = Context.make_context_id(Time.now(), node_id)
+    true = ContextId.valid?(context_id)
```
aMasakiTakahashi, line 38: I think it is better to define `make_context_id/2` function whose postcondition satisfies `ContextId.t`.
(To avoid checking the function output every time)

```elixir
defun make_context_id(t :: v[Time.t], node_id :: v[String.t]) :: v[ContextId.t] do
  context_id = [
    timestamp(t),
    node_id,
    :erlang.pid_to_list(self()) |> tl() |> List.to_string() |> String.trim_trailing(">"),
  ] |> Enum.join("_")
  true = ContextId.valid?(context_id)
  context_id
end
```

aYukiSekiguchi: It looks line `v[ContextId.t]` is enough to validate postcondition.
Fixed at https://github.com/access-company/antikythera/pull/126/commits/ce7ba1bcf65aa47d4d96f85b5b448742393616ed

lib/mix/task.ex
```diff
@@ -20,4 +24,17 @@ defmodule Antikythera.Mix.Task do
...
+
+  @doc """
+  Set the specified `node_id` to the `Antikythera.ContextId` in GearLog.
+
+  If you want to use GearLog in mix task, you must set `node_id` before you call GearLog functions.
+
+  The `Antikythera.ContextId` in GearLog will become `{timestamp}_{node_id}_{PID}`.
+  The `timestamp` and `PID` are automatically got from system.
+  """
+  defun set_node_id_to_gear_log_context(node_id :: v[String.t]) :: :ok do
```
ttokit, line 36: I'm concerned about the fact that I can't read the node_id constraints just by looking at this definition and documentation.
Can't we define a regular expression that can be specified as follows?
```ex
defmodule NodeId do
  use Croma.SubtypeOfString, pattern: ~R/\A[0-9A-Za-z.-]+\z/
end

defun set_node_id_to_gear_log_context(node_id :: v[NodeId.t]) :: :ok do
```

aYukiSekiguchi: Added `NodeId` in https://github.com/access-company/antikythera/pull/126/commits/cc3283e35cd770d0edd69bf701330da193f4ccd2

`AntikytheraCore.Cluster.NodeId.get/0` should return `~R/\A[0-9A-Za-z.-]+\z/`, but no one explicitly check it. I changed `AntikytheraCore.Cluster.NodeI` to use `Antikythera.NodeId` to verify it.  https://github.com/access-company/antikythera/pull/126/commits/30b24d726f96b681990a4ea6f2956e69951e0ad1

PR #128, opened 2020-05-28
[297359] Evaluate mix format

aMasakiTakahashi:
https://acsmine.tok.access-company.com/redmine/issues/207359


core/async_job/rafted_value_config_maker.ex
```diff
@@ -6,17 +6,17 @@ defmodule AntikytheraCore.AsyncJob.RaftedValueConfigMaker do
...
   @options [
-    heartbeat_timeout:                   1_000,
-    election_timeout:                    5_000,
-    election_timeout_clock_drift_margin:   500,
+    heartbeat_timeout: 1_000,
+    election_timeout: 5_000,
+    election_timeout_clock_drift_margin: 500
   ]
-  defun options() :: Keyword.t, do: @options
+  defun(options() :: Keyword.t(), do: @options)
```
ttokit, line 13: I hate...:-1:

core/alert/handler/handler.ex
```diff
@@ -36,7 +39,9 @@ defmodule AntikytheraCore.Alert.Handler do
   - Messages received by handlers will be sent out in "fast-then-delayed" pattern:
       - Occasional messages will be flushed from the buffer and sent out in `fast_interval`.
       - Messages arriving too frequently will be buffered for `delayed_interval` until they are sent out.
-  - By default, `fast_interval` is #{@default_fast_interval} seconds and `delayed_interval` is #{@default_delayed_interval} seconds.
+  - By default, `fast_interval` is #{@default_fast_interval} seconds and `delayed_interval` is #{
+    @default_delayed_interval
```
aYukiSekiguchi, line 43: Is the reason of this multi line that this line is too long?

core/async_job/rafted_value_config_maker.ex
aYukiSekiguchi: Is this problem because we use `defun` instead of `fun`?

ttokit: Yes, it looks like it can be worked around by using `def` or using a `do/end` block.

aMasakiTakahashi: Maybe so.
The following is the result of `mix format`.
* from
```
use Croma

defmodule Foo do
  @options [1]
  defun options() :: Keyword.t, do: @options
end

defmodule Bar do
  @options [1]
  def options() :: Keyword.t, do: @options
end
```
* diff
```diff
 defmodule Foo do
   @options [1]
-  defun options() :: Keyword.t, do: @options
+  defun(options() :: Keyword.t(), do: @options)
 end

 defmodule Bar do
   @options [1]
-  def options() :: Keyword.t, do: @options
+  def options() :: Keyword.t(), do: @options
 end
```

core/metrics/buffer.ex
```diff
@@ -17,33 +17,45 @@ defmodule AntikytheraCore.Metrics.Buffer do
   """
 
   alias Antikythera.Time
-  alias Antikythera.ExecutorPool.Id           , as: EPoolId
+  alias Antikythera.ExecutorPool.Id, as: EPoolId
```
aYukiSekiguchi, line 20: 👍

core/sup_tree_core/executor_pool/executor_pool.ex
```diff
@@ -7,89 +7,118 @@ defmodule AntikytheraCore.ExecutorPool do
...
-  alias AntikytheraCore.ExecutorPool.{ActionRunner, AsyncJobRunner, TimedJobStarter, WebsocketConnectionsCounter, UsageReporter, MemcacheWriter}
+
+  alias AntikytheraCore.ExecutorPool.{
+    ActionRunner,
+    AsyncJobRunner,
+    TimedJobStarter,
+    WebsocketConnectionsCounter,
+    UsageReporter,
+    MemcacheWriter
+  }
```
aYukiSekiguchi, line 18: I guess that the reason why this becomes multi lines is the original line is too long.
I don't think both of too long alias and `{}` multiline alias is good.
Then, if we see this type of alias, we should use single aliases.

aMasakiTakahashi: >I guess that the reason why this becomes multi lines is the original line is too long.

I also think so.
According to [here](https://github.com/elixir-lang/elixir/blob/4fea9cc942f632d5a1692eb85774d097546bc4e5/lib/elixir/lib/code.ex#L426) , `{}` seems to be the only place in which `formatter` can introduce a new line safely.

core/alert/handler/handler.ex
aMasakiTakahashi: >I guess that the reason why this becomes multi lines is the original line is too long.

I think so.
According to [here](https://github.com/elixir-lang/elixir/blob/4fea9cc942f632d5a1692eb85774d097546bc4e5/lib/elixir/lib/code.ex#L426) , `{}` seems to be the only place in which `formatter` can introduce a new line safely.

core/async_job/rafted_value_config_maker.ex
aDaichiOta: Does `:locals_without_parens` option solve this?

.formatter.exs
```diff
@@ -0,0 +1,3 @@
+[
+  inputs: ["{mix,.formatter}.exs", "{core,eal,lib,local,test}/**/*.{ex,exs}"]
```
aDaichiOta, line 2: This format has too many line breaks for me. I prefer larger `:line_length`.

aMasakiTakahashi: Do you have a recommended value as a maximum line length?

aDaichiOta: 100 or 120 ? I think it is too large if it is larger than 120.

core/async_job/rafted_value_config_maker.ex
aMasakiTakahashi: `:locals_without_parens` works well!
`diff` is the following.
```diff
 defmodule Foo do
   @options [heartbeat_timeout: 1_000]
-  defun options() :: Keyword.t, do: @options
+  defun options() :: Keyword.t(), do: @options
 end

 defmodule Bar do
   @options [heartbeat_timeout: 1_000]
-  def options() :: Keyword.t, do: @options
+  def options() :: Keyword.t(), do: @options
 end
```

.formatter.exs
aYukiSekiguchi: The default value of `:line_length` is [98](https://github.com/elixir-lang/elixir/blob/4fea9cc942f632d5a1692eb85774d097546bc4e5/lib/elixir/lib/code.ex#L653) which elixir style guide [recommends](https://github.com/christopheradams/elixir_style_guide#whitespace).

IMO, the current line length of Antikythera is too long for me who uses 12-13inch MacBook with vertically splitted editor.

I'd like to use 98.

core/alert/handler/handler.ex
irisTa56: I prefer enforcing developers to write document lines shorter than the threshold length rather than accepting such a line break. (at least, lines including interpolation)

core/core.ex
```diff
@@ -59,68 +66,88 @@ defmodule AntikytheraCore do
...
     }
+
     cowboy_proto_opts = %{
       # (a) timeout of a request with no data transfer; must be sufficiently longer than the gear action timeout (10_000)
       # (b) timeout of a connection with no requests; this should be longer than LB's idle timeout
-      idle_timeout:    30_000,  # (a)
-      request_timeout: 120_000, # (b)
-      env:             %{dispatch: dispatch_rules},
-      stream_handlers: [:cowboy_compress_h, :cowboy_stream_h],
+      # (a)
```
irisTa56, line 95: When formatted, I prefer directly inserting comment like
```suggestion
      # timeout of a request with no data transfer; must be sufficiently longer than the gear action timeout (10_000)
```

aMasakiTakahashi:
Thank you for your feedback.
I checked all comments and summarized them below.

* [`defun` is regarded as a function](https://github.com/access-company/antikythera/pull/128#discussion_r431627818)
  * `:locals_without_parens` seems to work well.
* [Weird linebreak policy](https://github.com/access-company/antikythera/pull/128#discussion_r431629027)
  * Making a rule not to write too long document lines, we can control automatic linebreak to a certain extent.
* [Multiline alias](https://github.com/access-company/antikythera/pull/128#discussion_r431643845)
  * We can avoid this by using a single alias.
* [Comment behind a code](https://github.com/access-company/antikythera/pull/128#discussion_r433197742)
  * Depends on how to comment (, so we can control).
* [Appropriate line length](https://github.com/access-company/antikythera/pull/128#discussion_r432209061)
  * We can discuss this separately from whether `mix format` should be introduced.

There seems to be some kind of workaround to all of the issues.
If there are other issues or insufficient workarounds, please add a comment.

aMasakiTakahashi:
Thanks to your review, it has become clear that there are a few things that need to be changed before formatting.
I'll fix it in the new PR, not in this PR.
(Since `mix format` changes quite a few codes, it's hard to distinguish them from the automatic changes in the code.)

PR #130, opened 2020-06-17
[206806] Allow access via localhost

aDaichiOta:
We no longer need to modify `/etc/hosts` for gear development.

https://acsmine.tok.access-company.com/redmine/issues/206806#note-5

### DIFF
--- a/core/handler/cowboy_routing.ex
+++ b/core/handler/cowboy_routing.ex
@@ -30,32 +30,35 @@ defmodule AntikytheraCore.Handler.CowboyRouting do
       @total_error_count_route,
       @per_app_error_count_route,
     ]
-    [{:_, default_gear_routes() ++ path_rules}]
+    case default_routing_gear() do
+      nil  -> [{:_, path_rules}]
+      gear -> [{:_, gear_routes(gear) ++ path_rules}]
+    end
   end
 
-  defunp default_gear_routes() :: [route_path] do
+  defunp default_routing_gear() :: GearName.t | nil do
     if :code.is_loaded(Mix.Project) do
       conf = Mix.Project.config()
-      if Keyword.get(conf, :antikythera_gear) != nil do
-        gear_name = Keyword.get(conf, :app)
-        [
-          static_file_serving_route(gear_name),
-          normal_routes(gear_name),
-        ] |> Enum.reject(&is_nil/1)
+      if conf[:antikythera_gear] != nil do
+        conf[:app]
       else
-        []
+        nil
       end
     else
-      []
+      nil
     end
   end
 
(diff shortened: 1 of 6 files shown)


core/handler/cowboy_routing.ex
```diff
@@ -28,7 +30,24 @@ defmodule AntikytheraCore.Handler.CowboyRouting do
       @total_error_count_route,
       @per_app_error_count_route,
     ]
-    [{:_, path_rules}]
+    [{:_, default_gear_routes() ++ path_rules}]
```
aYukiSekiguchi, line 33: I understand that this is WIP.
What happens if a gear has `/healtheck` path?
I don't think it should work, but we need some documentation about it.

aDaichiOta: `/healthcheck` returns the gear API instead of antikythera.
I think this behavior is ok because gear developers don't use antikythera's health check API.

Or do you mean the following implementation would be better?
It may be confusing for gear developers that APIs not written in `router.ex` return responses.
```elixir
[{:_, default_gear_routes()]
```

aYukiSekiguchi: > /healthcheck returns the gear API instead of antikythera.
> I think this behavior is ok because gear developers don't use antikythera's health check API.

SGTM.
I don't think we need a documentation if gear API overwrites Antikythera API

core/handler/cowboy_routing.ex
```diff
@@ -28,19 +30,37 @@ defmodule AntikytheraCore.Handler.CowboyRouting do
       @total_error_count_route,
       @per_app_error_count_route,
     ]
-    [{:_, path_rules}]
+    case default_routing_gear() do
+      :error      -> [{:_, path_rules}]
+      {:ok, gear} -> [{:_, gear_routes(gear) ++ path_rules}]
```
aYukiSekiguchi, line 35: As far as I tested on the current master on iex,
* `http://gearname.localhost:8080` works
* `http://gearname.myhostname:8080` doesn't work
* `http://myhostname:8080/healthcheck` works even if I accessed from an other PC

It looks like I can access my gear from an other PC using `http://myhostname:8080` url after this change.
Is this correct?

I think it's a bit insecure if other people can access my developing gear.

Can we limit this to `*.localhost` or `*.BASE_DOMAIN`?

aDaichiOta: Thanks. I'll fix it.

aDaichiOta: fixed. 7ac54ff7808f60fdd148c1903a7aaa07cd6eaf9d

PR #131, opened 2020-06-22
[209104] Add support for a custom policy to CloudfrontSignedUrl

aYukiSekiguchi:
https://acsmine.tok.access-company.com/redmine/issues/209104

### DIFF
--- a/lib/util/aws/cloudfront_signed_url.ex
+++ b/lib/util/aws/cloudfront_signed_url.ex
@@ -31,13 +31,13 @@ defmodule Antikythera.Aws.CloudfrontSignedUrl do
     encoded_url = if url_encoded?, do: resource_url, else: URI.encode(resource_url)
     expires_in_seconds = System.system_time(:second) + lifetime_in_seconds
     joiner = if URI.parse(encoded_url) |> Map.get(:query) |> is_nil(), do: "?", else: "&"
-    encoded_url <> joiner <> URI.encode_query(make_query_params(encoded_url, expires_in_seconds, key_pair_id, private_key))
+    encoded_url <> joiner <> URI.encode_query(make_query_params_for_canned_policy(encoded_url, expires_in_seconds, key_pair_id, private_key))
   end
 
-  defunpt make_query_params(encoded_url        :: v[String.t],
-                            expires_in_seconds :: v[pos_integer],
-                            key_pair_id        :: v[String.t],
-                            private_key        :: v[String.t]) :: [{String.t, String.t}] do
+  defunpt make_query_params_for_canned_policy(encoded_url        :: v[String.t],
+                                              expires_in_seconds :: v[pos_integer],
+                                              key_pair_id        :: v[String.t],
+                                              private_key        :: v[String.t]) :: [{String.t, String.t}] do
     policy_statement =
       ~s/{"Statement":[{"Resource":"#{encoded_url}","Condition":{"DateLessThan":{"AWS:EpochTime":#{expires_in_seconds}}}}]}/
     signature = create_signature(policy_statement, private_key)
@@ -48,14 +48,19 @@ defmodule Antikythera.Aws.CloudfrontSignedUrl do
     ]
   end
 
-  defunp create_signature(policy_statement :: v[String.t], private_key :: v[String.t]) :: String.t do
-    :public_key.sign(policy_statement, :sha, decode_rsa_key(private_key))
+  defunp encode_for_aws(string :: v[String.t]) :: v[String.t] do
+    string
     |> Base.encode64()
     # Replace characters that are invalid in a URL query string with characters that are valid.
     # https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-canned-policy.html
     |> String.replace("+", "-")
     |> String.replace("=", "_")
     |> String.replace("/", "~")
+  end
+
+  defunp create_signature(policy_statement :: v[String.t], private_key :: v[String.t]) :: String.t do
+    :public_key.sign(policy_statement, :sha, decode_rsa_key(private_key))
+    |> encode_for_aws()
   end
 
   defp decode_rsa_key(rsa_key) do
--- a/lib/util/aws/cloudfront_signed_url.ex
+++ b/lib/util/aws/cloudfront_signed_url.ex
@@ -4,7 +4,7 @@ use Croma
 
 defmodule Antikythera.Aws.CloudfrontSignedUrl do
   @moduledoc """
-  This module provides `generate_signed_url/5` to generate [a signed URL for CloudFront](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-signed-urls.html)
+  This module provides functions to generate [a signed URL for CloudFront](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-signed-urls.html)
   """
 
   @doc """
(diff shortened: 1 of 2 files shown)


lib/util/aws/cloudfront_signed_url.ex
```diff
@@ -31,13 +30,43 @@ defmodule Antikythera.Aws.CloudfrontSignedUrl do
...
+
+  ## Parameters
+
+  - `resource_url` (string): CloudFront URL used for accessing the file, including query string parameters, if any.
+  - `lifetime_in_seconds` (positive integer): Expiration time is determined as the sum of the current time (in seconds) and this value.
+  - `key_pair_id` (string): ID for an active CloudFront key pair used for generating the signature.
+  - `private_key` (string): RSA private key for the key pair specified by `key_pair_id`.
+  - `url_encoded?` (boolean): Whether `resource_url` is encoded or not (optional, default is `false`).
+  - `optional_policy` (Keyword): Optional policy conditions to be added to a custom policy (default is `[]`). Currently, supports only the following keywords:
+    - `:date_greater_than`(integer >= 0): Seconds from `AWS:EpockTime`. Specified to `DateGreaterThan`
```
irisTa56, line 47: Does `AWS:EpockTime` have the same meaning to the Unix epoch?
If not, I suggest:
```suggestion
    - `:date_greater_than` (integer >= 0): Seconds from the Unix epoch. Specified to `DateGreaterThan`.
```

aYukiSekiguchi: CloudFront document [says](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html#private-content-custom-policy-statement)
> date and time in Unix time format and UTC

Then, this should be the Unix epoch.

However, I'm not sure what will happen after [Y2038](https://en.wikipedia.org/wiki/Year_2038_problem).

Since this module is a thin wrapper for CloudFront, a user should know about CloudFront. It would be better to delegate the correct definition to AWS' document.

aYukiSekiguchi: I'll fix this typo.

```suggestion
    - `:date_greater_than`(integer >= 0): Seconds from `AWS:EpochTime`. Specified to `DateGreaterThan`
```

irisTa56: Thank you for your explanation!

There might be misunderstanding what I meant by "the Unix epoch".
I used "the Unix epoch" as a specific time point (1970-01-01 00:00:00 UTC), not a time specification.

Since `AWS:EpockTime` is a time specification, I think changing "Seconds from `AWS:EpochTime`" to "Seconds in `AWS:EpochTime`" can avoid such confusion.

aYukiSekiguchi: Fixed at https://github.com/access-company/antikythera/pull/131/commits/d25bcc10ce1ab908e224886d3f4d55df36ae7744

aYukiSekiguchi:
I changed `:ip_address` from `string` to `a list of strings` at https://github.com/access-company/antikythera/pull/131/commits/bbcbf09ea8de12e63900d40762d58d31538bb3d3.

@aMasakiTakahashi @aDaichiOta 
Could you review the commit?

lib/util/aws/cloudfront_signed_url.ex
```diff
@@ -48,8 +77,47 @@ defmodule Antikythera.Aws.CloudfrontSignedUrl do
...
+        {:ok, t} when is_integer(t) and t >= 0 -> ~s/,"DateGreaterThan":{"AWS:EpochTime":#{t}}/
+        _                                      -> ""
+      end
+    ip_address =
+      case Keyword.fetch(optional_policy, :ip_address) do
+        {:ok, []}        -> ""
+        {:ok, addresses} ->
+          # It is important to restrict a connection using IpAddress, so this raises a MatchError instead of just ignoring :ip_address.
+          true = is_list(addresses)
+          ~s/,"IpAddress":{"AWS:SourceIp":#{generate_ip_address_string(addresses)}}/
```
aDaichiOta, line 99: Maybe we can use `when`
```suggestion
        {:ok, addresses} when is_list(addresses) -> ~s/,"IpAddress":{"AWS:SourceIp":#{generate_ip_address_string(addresses)}}/          
```

lib/util/aws/cloudfront_signed_url.ex
```diff
@@ -48,8 +77,47 @@ defmodule Antikythera.Aws.CloudfrontSignedUrl do
     ]
   end
 
-  defunp create_signature(policy_statement :: v[String.t], private_key :: v[String.t]) :: String.t do
-    :public_key.sign(policy_statement, :sha, decode_rsa_key(private_key))
+  defunpt generate_ip_address_string(addresses :: v[[String.t]]) :: v[String.t] do
+    case addresses do
+      [address]     -> ~s/"#{String.trim(address)}"/
+      [address | t] -> Enum.reduce(t, ~s/["#{String.trim(address)}"/, fn (address, result) -> result <> ~s/,"#{String.trim(address)}"/ end) <> "]"
```
aDaichiOta, line 83: (optional)
When this function is called, `addresses` aren't empty, so you could write:
```elixir
case addresses do
  [address] -> ~s/"#{String.trim(address)}"/
  addresses -> "[" <> Enum.map_join(addresses, ",", fn address -> ~s/"#{String.trim(address)}"/ end) <> "]"
end
```

lib/util/aws/cloudfront_signed_url.ex
```diff
@@ -48,8 +77,47 @@ defmodule Antikythera.Aws.CloudfrontSignedUrl do
...
+      case Keyword.fetch(optional_policy, :date_greater_than) do
+        {:ok, t} when is_integer(t) and t >= 0 -> ~s/,"DateGreaterThan":{"AWS:EpochTime":#{t}}/
+        _                                      -> ""
+      end
+    ip_address =
+      case Keyword.fetch(optional_policy, :ip_address) do
+        {:ok, []}        -> ""
+        {:ok, addresses} ->
+          # It is important to restrict a connection using IpAddress, so this raises a MatchError instead of just ignoring :ip_address.
+          true = is_list(addresses)
```
aMasakiTakahashi, line 98: I would like to check the type of `addresses` with `guard` for several reasons.
* The reason for the exception is more obvious.
* We can separate the success case and error case clearly.
```
case Keyword.fetch(optional_policy, :ip_address) do
  {:ok, []}                                -> ""
  {:ok, addresses} when is_list(addresses) -> ~s/,"IpAddress":{"AWS:SourceIp":#{generate_ip_address_string(addresses)}}/
  # It is important to restrict a connection using IpAddress, so this raises a RuntimeError instead of just ignoring :ip_address.
  {:ok, _}                                 -> raise "Type of ip_address is not list"
  _                                        -> ""
end
```

lib/util/aws/cloudfront_signed_url.ex
aYukiSekiguchi: It's better than the current code. Fixed at https://github.com/access-company/antikythera/pull/131/commits/0fee513181c93d0d4971c6df9736e4ec3ee2c0dd

lib/util/aws/cloudfront_signed_url.ex
aYukiSekiguchi: Fixed at https://github.com/access-company/antikythera/pull/131/commits/3ed9433cc5f0533b7a645fd77adb8694e6e16e4a

PR #152, opened 2020-11-19
[216773] Throw MatchError if about half of gears are not installed

aYukiSekiguchi:
https://acsmine.tok.access-company.com/redmine/issues/216773

Antikythera returns healthy to healthy API even if there is no running gear.  Generally speaking, this in not healthy.

`Version.Gear.install_gears_at_startup/1` returns :error if equal or less than half of gears are installed. Then, `AntikytheraCore.VersionSynchronizer.install_gears/0` try to match it with `:ok`. This won't call `StartupManager.all_gears_installed/0` which makes Antikythera healthy state.

### DIFF
--- a/core/version/gear.ex
+++ b/core/version/gear.ex
@@ -157,7 +157,7 @@ defmodule AntikytheraCore.Version.Gear do
     needs_check = Antikythera.Env.runtime_env() != :local
 
     is_enough_gears_installed =
-      num_installable != 0 && num_installed / num_installable > @installed_gear_ratio_threshold
+      num_installable != 0 and num_installed / num_installable > @installed_gear_ratio_threshold
 
     if !needs_check || is_enough_gears_installed do
       :ok
--- a/core/version/gear.ex
+++ b/core/version/gear.ex
@@ -130,21 +130,28 @@ defmodule AntikytheraCore.Version.Gear do
         {g, __MODULE__.gear_dependencies_from_app_file(g, gear_names)}
       end)
 
-    pairs_not_installed =
-      __MODULE__.install_gears_whose_deps_met(gear_and_deps_pairs, MapSet.new(), fn gear_name ->
-        try do
-          install_or_upgrade_to_next_version(gear_name)
-        rescue
-          e -> L.error("Failed to install #{gear_name}: #{Exception.message(e)}")
+    {pairs_not_installed, num_failed_install} =
+      __MODULE__.install_gears_whose_deps_met(
+        gear_and_deps_pairs,
+        MapSet.new(),
+        0,
+        fn gear_name ->
+          try do
+            install_or_upgrade_to_next_version(gear_name)
+          rescue
+            e ->
+              L.error("Failed to install #{gear_name}: #{Exception.message(e)}")
+              :error
+          end
         end
-      end)
+      )
 
     Enum.each(pairs_not_installed, fn {gear_name, deps} ->
       L.error("#{gear_name} is not installed due to unmatched dependencies: #{inspect(deps)}")
     end)
 
     num_installable = length(gear_names)
-    num_installed = num_installable - length(pairs_not_installed)
+    num_installed = num_installable - length(pairs_not_installed) - num_failed_install
     # version_upgrade_test starts Antikythera without gears.
     # Then, we can't check number of installed gears.
     needs_check = Antikythera.Env.runtime_env() != :local
@@ -163,24 +170,26 @@ defmodule AntikytheraCore.Version.Gear do
   defun install_gears_whose_deps_met(
           pairs :: v[[gear_and_deps_pair]],
           installed_gears_set :: MapSet.t(),
-          f :: (GearName.t() -> :ok)
-        ) :: [gear_and_deps_pair] do
+          num_failed_install :: v[non_neg_integer],
+          f :: (GearName.t() -> :ok | :error)
+        ) :: {[gear_and_deps_pair], non_neg_integer} do
     if Enum.empty?(pairs) do
-      []
+      {[], num_failed_install}
     else
       {pairs_installable, pairs_not_installable} =
         Enum.split_with(pairs, fn {_, deps} -> MapSet.subset?(deps, installed_gears_set) end)
 
       if Enum.empty?(pairs_installable) do
         # we cannot make progress any more
-        pairs_not_installable
+        {pairs_not_installable, num_failed_install}
       else
         gears_installable = Keyword.keys(pairs_installable)
-        Enum.each(gears_installable, f)
+        gears_installed = Enum.filter(gears_installable, fn gear -> f.(gear) == :ok end)
 
         install_gears_whose_deps_met(
           pairs_not_installable,
-          Enum.into(gears_installable, installed_gears_set),
+          Enum.into(gears_installed, installed_gears_set),
+          num_failed_install + length(gears_installable) - length(gears_installed),
           f
         )
       end
--- a/core/version/gear.ex
+++ b/core/version/gear.ex
@@ -9,6 +9,8 @@ defmodule AntikytheraCore.Version.Gear do
   alias AntikytheraCore.Path, as: CorePath
   require AntikytheraCore.Logger, as: L
 
+  @installed_gear_ratio_threshold 0.5
+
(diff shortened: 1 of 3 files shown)


core/version/gear.ex
```diff
@@ -113,20 +115,23 @@ defmodule AntikytheraCore.Version.Gear do
...
+    # Tests need `__MODULE__.` to mock these functions.
     gear_and_deps_pairs =
-      Enum.map(gear_names, fn g -> {g, gear_dependencies_from_app_file(g, gear_names)} end)
+      Enum.map(gear_names, fn g ->
+        {g, __MODULE__.gear_dependencies_from_app_file(g, gear_names)}
+      end)
 
-    pairs_not_installed =
-      install_gears_whose_deps_met(gear_and_deps_pairs, MapSet.new(), fn gear_name ->
+    {pairs_not_installed, num_installed} =
```
aDaichiOta, line 133: I think the value of num_installed can be calculated as follows:
```elixir
num_installed = length(gear_names) - length(pairs_not_installed)
```

core/version/gear.ex
```diff
@@ -137,22 +142,37 @@ defmodule AntikytheraCore.Version.Gear do
...
       {pairs_installable, pairs_not_installable} =
         Enum.split_with(pairs, fn {_, deps} -> MapSet.subset?(deps, installed_gears_set) end)
 
       if Enum.empty?(pairs_installable) do
         # we cannot make progress any more
-        pairs_not_installable
+        {pairs_not_installable, MapSet.size(installed_gears_set)}
       else
         gears_installable = Keyword.keys(pairs_installable)
         Enum.each(gears_installable, f)
```
aDaichiOta, line 178: There are two possible reasons why the gear can't be installed. One is the circular reference between gears, the other is the problem of the gear application itself.
This code counts the former problem, but not the latter. We need to check the return value of `f`.

core/version/gear.ex
aYukiSekiguchi: Fixed at
https://github.com/access-company/antikythera/pull/152/commits/a97aef3a7f5b8e49e6b083779fef95fb9288372c

core/version/gear.ex
aYukiSekiguchi: I think it should not consider the gears which are failed to install as installed gear, so I fixed this as well.
Since the gears which are failed to install are already reported when `install_gears_whose_deps_met/4` ends, `pairs_not_installable` isn't good variable to save the gears. `install_gears_whose_deps_met/4` returns the number of the gears. Then, `do_install_gears_at_startup/1` uses the number to decide whether it returns :ok.
https://github.com/access-company/antikythera/pull/152/commits/3298e5983cebb6670eaa34b725de484a90d819bb
https://github.com/access-company/antikythera/pull/152/commits/4c5a4d2181c8e42e029b56edd4ca5eccdb785d43

core/version/gear.ex
```diff
@@ -113,63 +115,92 @@ defmodule AntikytheraCore.Version.Gear do
...
-      []
+      {[], num_failed_install}
     else
       {pairs_installable, pairs_not_installable} =
         Enum.split_with(pairs, fn {_, deps} -> MapSet.subset?(deps, installed_gears_set) end)
 
       if Enum.empty?(pairs_installable) do
         # we cannot make progress any more
-        pairs_not_installable
+        {pairs_not_installable, num_failed_install}
```
aDaichiOta, line 184: **(optional)**

If the following error log is written here, the return value of this function will be simple (the number of gears that could not be installed).

```elixir
    Enum.each(pairs_not_installed, fn {gear_name, deps} ->
      L.error("#{gear_name} is not installed due to unmatched dependencies: #{inspect(deps)}")
    end)
```

test/core/version/gear_test.exs
```diff
@@ -15,30 +15,188 @@ defmodule AntikytheraCore.Version.GearTest do
...
+      pairs = [gear_a: set([]), gear_b: set([]), gear_c: set([])]
+      failed_to_install_gear = :gear_b
+
+      {ret, num_failed_install} =
+        V.install_gears_whose_deps_met(pairs, MapSet.new(), 0, fn g ->
+          if g == failed_to_install_gear do
+            :error
+          else
+            :ok
+          end
```
aDaichiOta, line 64: **(optional)**

```suggestion
          if g == failed_to_install_gear, do: :error, else: :ok
```

core/version/gear.ex
```diff
@@ -113,63 +115,92 @@ defmodule AntikytheraCore.Version.Gear do
...
+
     Enum.each(pairs_not_installed, fn {gear_name, deps} ->
       L.error("#{gear_name} is not installed due to unmatched dependencies: #{inspect(deps)}")
     end)
+
+    num_installable = length(gear_names)
+    num_installed = num_installable - length(pairs_not_installed) - num_failed_install
+    # version_upgrade_test starts Antikythera without gears.
+    # Then, we can't check number of installed gears.
+    needs_check = Antikythera.Env.runtime_env() != :local
```
aDaichiOta, line 157: `is_enough_gears_installed` seems to handle even if the gear is empty. Isn't `needs_check` unnecessary?

aYukiSekiguchi: Sorry if I misunderstand your question.

If `num_installable` is 0, `is_enough_gears_installed` becomes `false` and this returns `:error`.
Since `install_gears/0` matches this return value with `:ok`, Antikythera initialization will throw MatchError.
Since Antiktyhera isn't initialized, `:rpc.call/4` in `Mix.Tasks.AntikytheraLocal.VersionUpgradeTest`, which run on `ANTIKYTHERA_COMPILE_ENV=local` , will throw an exception.
 
`needs_check` is needed to make this function return `:ok` even if `num_installable == 0`.

core/version/gear.ex
aYukiSekiguchi:
```
the current code may be more appropriate because both error loggings are in do_install_gears_at_startup/1.
```

I think this is appropriate.
`install_gears_whose_deps_met/4` is something like iterator, and `do_install_gears_at_startup/1` does the actual work i.e. installing gears and reporting errors.

core/version/gear.ex
aDaichiOta: Thaks. I read the code incorrectly.

aDaichiOta: @Fumipo-Theta
I have a question about the internal environment. Is there possibility that there are no gears when starting Antikythera?

test/core/version/gear_test.exs
aYukiSekiguchi: Fixed at https://github.com/access-company/antikythera/pull/152/commits/659ba110cc1b9f0da98b82aa88d74f0a9a2e5cdb

core/version/gear.ex
Fumipo-Theta: In the internal environment, Erlang nodes start after the artifacts of two gears are generated and Antikythera release is made.
So, I think there should be at least two gears when starting Antikythera.
I made sure Antikythera starts with the two gears.

aDaichiOta (approved):
LGTM

Fumipo-Theta (approved):
LGTM !

core/version/gear.ex
```diff
@@ -113,63 +115,92 @@ defmodule AntikytheraCore.Version.Gear do
...
+
+    is_enough_gears_installed =
+      num_installable != 0 && num_installed / num_installable > @installed_gear_ratio_threshold
```
aMasakiTakahashi, line 160: I think `and` operator is better if both sides of the operator are expected to be boolean.
```suggestion
      num_installable != 0 and num_installed / num_installable > @installed_gear_ratio_threshold
```

core/version/gear.ex
aYukiSekiguchi: Thank you for the description in public place.

Note: Internal Slack discussion about it:
https://accessjp.slack.com/archives/CDEN4MANQ/p1605694056455100

core/version/gear.ex
aYukiSekiguchi: Fixed at https://github.com/access-company/antikythera/pull/152/commits/2473a91ba577ab05ec7b6ba9607577074ebe4d01

aMasakiTakahashi (approved):
LGTM!

PR #165, opened 2021-02-02
[222347] Change parameters to handle larger gear log size

Fumipo-Theta:
https://acsmine.tok.access-company.com/redmine/issues/222347#note-8

If there are hundreds of thousands of requests per minute to a gear, the size of the log file increase several tens of MB and may exceed the max body size limit.
So,
- I suggest increasing the max body size limit.
- I also suggest shortening the size check interval to reduce the rapid increase in log file size.


aYukiSekiguchi (approved):
LGTM

aMasakiTakahashi (approved):
LGTM!

aDaichiOta (approved):
LGTM

ttokit (approved):

PR #167, opened 2021-02-19
[220238] Add `Antikythera.Env.base_url/1`

aDaichiOta:
https://gate.tok.access-company.com/redmine/issues/220238

### DIFF
--- a/lib/util/env.ex
+++ b/lib/util/env.ex
@@ -141,6 +141,31 @@ defmodule Antikythera.Env do
     end
   end
 
+  if Antikythera.Env.Mapping.cloud?(@compile_env) do
+    @scheme "https"
+  else
+    @scheme "http"
+  end
+
+  @doc """
+  Return the base URL which based on the `Host` HTTP header of the request.
+
+  This function is useful if you use a custom domain.
+  Whether the scheme of the URL is `https` or `http` depends on whether `Antikythera.Env.Mapping.cloud?/1` returns true or false.
+  """
+  defun base_url(%Antikythera.Conn{request: %Antikythera.Request{headers: headers}}) :: v[Url.t()] do
+    case headers do
+      %{"host" => host} -> @scheme <> "://" <> host
+      _ -> raise "`Host` header is not in the request"
+    end
+  end
+
+  @doc """
+  Return the base URL of the gear.
+
+  If the gear name is `my_gear` and the Antikythera is deployed at `antikythera.example.com`,
+  this function returns `https://my-gear.antikythera.example.com`.
+  """
   defun default_base_url(
           gear_name :: v[GearName.t() | GearNameStr.t()],
           env :: v[t] \\ @compile_env
--- a/lib/util/env.ex
+++ b/lib/util/env.ex
@@ -154,9 +154,9 @@ defmodule Antikythera.Env do
   Whether the scheme of the URL is `https` or `http` depends on whether `Antikythera.Env.Mapping.cloud?/1` returns true or false.
   """
   defun base_url(%Antikythera.Conn{request: %Antikythera.Request{headers: headers}}) :: v[Url.t()] do
-    case headers do
-      %{"host" => host} -> @scheme <> "://" <> host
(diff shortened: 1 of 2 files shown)


lib/util/env.ex
```diff
@@ -141,6 +141,31 @@ defmodule Antikythera.Env do
     end
   end
 
+  if Antikythera.Env.Mapping.cloud?(@compile_env) do
+    @scheme "https"
```
aYukiSekiguchi, line 145: Optional:
It looks like Antikythera isn't support `http` on cloud for both of Antikythera's domain and a custom domain of gears.
It would be better if `default_base_url/2` uses this to share the above knowledge.

aYukiSekiguchi (approved):
LGTM

lib/util/env.ex
aDaichiOta: `default_base_url/2` takes `env` as the second argument. Do you mean that the following code is better?

```elixir
@doc """
...
Whether the scheme of the URL is `https` or `http` depends on whether `Antikythera.Env.Mapping.cloud?/1` returns true or false.
"""
def base_url(conn)

@doc """
...
Whether the scheme of the URL is `https` or `http` depends on whether `Antikythera.Env.Mapping.cloud?/1` returns true or false.
"""
def default_base_url(gear_name), do: default_base_url(gear_name, @compile_env)

@doc false
def default_base_url(gear_name, env)
```

Fumipo-Theta: I suppose use cases of `default_base_url/1` and `default_base_url/2` are different.
The former is used in the running gear as same as `base_url/1`, and the latter may be used from outside the gear (connecting to the deployed gear from local for example, so we need to indicate `env` explicitly).
Therefore I think separating interface like above example would be good.

aYukiSekiguchi: Sorry, I missed your comment.

I meant that we should use `@scheme` in default_base_url as well:
```elixir
  defun default_base_url(
          gear_name :: v[GearName.t() | GearNameStr.t()],
          env :: v[t] \\ @compile_env
        ) :: Url.t() do
    if Mapping.cloud?(env) do
      @scheme <> "://" <> Routing.default_domain(gear_name, env)
    else
      @scheme <> "://" <> Routing.default_domain(gear_name, env) <> ":#{port_to_listen()}"
    end
  end
```

Since we always add `"://"` to `@scheme`, you can change `@scheme` to `"https://"` and its name to `@base_url_prefx` or something.

aDaichiOta: One gear uses `default_base_url(gear_name, env)` to get the deployment environment URL from local mix task.
So it is necessary to decide the scheme based on the `env` argument.

lib/util/env.ex
```diff
@@ -141,6 +141,31 @@ defmodule Antikythera.Env do
...
+  Return the base URL which based on the `Host` HTTP header of the request.
+
+  This function is useful if you use a custom domain.
+  Whether the scheme of the URL is `https` or `http` depends on whether `Antikythera.Env.Mapping.cloud?/1` returns true or false.
+  """
+  defun base_url(%Antikythera.Conn{request: %Antikythera.Request{headers: headers}}) :: v[Url.t()] do
+    case headers do
+      %{"host" => host} -> @scheme <> "://" <> host
+      _ -> raise "`Host` header is not in the request"
+    end
```
ttokit, line 160: (optional)
```suggestion
    case Map.get(headers, "host") do
      nil -> raise "`Host` header is not in the request"
      host -> @scheme <> "://" <> host
    end
```

ttokit (approved):

lib/util/env.ex
aYukiSekiguchi: Ah, sorry. I misunderstood `default_base_url/2`.
You're right!

aYukiSekiguchi (approved):
LGTM

Fumipo-Theta (approved):
LGTM !
Thank you for adding example usage and test cases.

PR #171, opened 2021-06-08
[222104] Add a new field to gear config

irisTa56:
https://acsmine.tok.access-company.com/redmine/issues/222104#note-21

### DIFF
--- a/core/config/gear.ex
+++ b/core/config/gear.ex
@@ -27,7 +27,10 @@ defmodule AntikytheraCore.Config.Gear do
       kv: Croma.Map,
       domains: CustomDomainList,
       log_level: Level,
-      alerts: HandlerConfigsMap
+      alerts: HandlerConfigsMap,
+      # can be used to store instance-specific information
+      # whose structure needs to be managed by administrative gears
+      internal_kv: {Croma.Map, default: %{}}
     ]
 
   defun default() :: t do
--- a/test/core/config/gear_test.exs
+++ b/test/core/config/gear_test.exs
@@ -3,17 +3,27 @@
 defmodule AntikytheraCore.Config.GearTest do
   use Croma.TestCase
   alias Antikythera.Test.GenServerHelper
+  alias Antikythera.Crypto.Aes
+  alias AntikytheraCore.Path, as: CorePath
   alias AntikytheraCore.GearLog.Level
   alias AntikytheraCore.Ets.ConfigCache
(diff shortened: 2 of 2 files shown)


aMasakiTakahashi (approved):
LGTM!

Fumipo-Theta (approved):
LGTMです。

aDaichiOta (approved):

core/config/gear.ex
```diff
@@ -27,7 +27,10 @@ defmodule AntikytheraCore.Config.Gear do
       kv: Croma.Map,
       domains: CustomDomainList,
       log_level: Level,
-      alerts: HandlerConfigsMap
+      alerts: HandlerConfigsMap,
+      # can be used to store instance-specific information
+      # whose structure needs to be managed by administrative gears
+      internal_kv: {Croma.Map, default: %{}}
```
aYukiSekiguchi, line 33: IIUC, since this struct is used like `%__MODULE__{kv: %{}...}` in `default/0`, this `default: %{}` isn't used.
It would be better to add `internal_kv: %{}` to `default/0`

aYukiSekiguchi (approved):
LGTM

core/config/gear.ex
irisTa56: This `default: %{}` is required when reading a config which is written in the current format.
Since a config in an existing file lack "internal_kv" field, `new/1` in line 61 will cause an error when reading the config.
`default/0` will be called only when a config file itself doesn't exist.

aYukiSekiguchi: Do you mean the following?
* If `read/1` is called for an existing gear, `read/1` returns `internal_kv: %{}` with the other keys because `default: %{}` in this croma struct is used.
* If `read/1` is called for a new gear, `read/1` returns the other keys without `internal_kv: %{}` because `default/0` is called.

irisTa56: > * If `read/1` is called for an existing gear, `read/1` returns `internal_kv: %{}` with the other keys because `default: %{}` in this croma struct is used.

Yes, I mean it.

> * If `read/1` is called for a new gear, `read/1` returns the other keys without `internal_kv: %{}` because `default/0` is called.

No.
In this case, `internal_kv: %{}` is also included.
Croma struct implicitly do this.

aYukiSekiguchi: Understand!
Thank you for clarify.

PR #172, opened 2021-06-22
[228684] gear config cache should not return core config

irisTa56:
https://acsmine.tok.access-company.com/redmine/issues/228684#note-7


core/ets/config_cache.ex
```diff
@@ -30,10 +30,11 @@ defmodule AntikytheraCore.Ets.ConfigCache.Gear do
   alias Antikythera.GearName
   alias AntikytheraCore.Config.Gear, as: GearConfig
 
-  defun read(gear_name :: v[GearName.t()]) :: nil | GearConfig.t() do
+  defun read(gear_name :: v[GearName.t()]) :: nil | GearConfig.t() | :error do
     case :ets.lookup(@table_name, gear_name) do
       [] -> nil
-      [{_gear_name, gear_config}] -> gear_config
+      [{_gear_name, %GearConfig{} = gear_config}] -> gear_config
+      _ -> :error
```
aYukiSekiguchi, line 37: I don't think this is needed. `CaseClauseError` is better.

Since Antikythera bans to create a gear whose name is `antikythera` and all code in Antikythera assume that `anitkythera` isn't passed to `GearName.t()`, this case should be critical.

If we add `:error`, someone can write code which sometimes pass `antikythera` to `GearName.t` and recover using this `:error`. This breaks the assumption.

Anyway, all code which use this `read/1` don't handle `:error` now, it will rise `CaseClauseError` or `BadMapError` if `antikythera` is passed.

irisTa56: > If we add `:error`, someone can write code which sometimes pass `antikythera` to `GearName.t` and recover using this `:error`. This breaks the assumption.

Thank you! That makes total sense.
fixed at https://github.com/access-company/antikythera/pull/172/commits/9311a5cf1ac53fd9685081b0f2874e99fb85198c

aYukiSekiguchi: Thank you for fixing.
However, it would be better to test error cases. You added the test case for `:antikythera`. It was great.

You can use `assert_raise/2`.
```elixir
    # normal behaviors are tested in AntikytheraCore.Config.GearTest
    test "should raise CaseClauseError if :antikythera is given as gear name" do
      assert_raise CaseClauseError, fn ->
        Gear.read(:antikythera)
      end
    end
```

irisTa56: Indeed, test clarify what we need to care about.
I borrow your code as is. Thank you!!
https://github.com/access-company/antikythera/pull/172/commits/d6e74f5c3f6d9e49c282966cce760f0a45debbc6

aYukiSekiguchi (approved):
LGTM

Fumipo-Theta (approved):
LGTM!

aMasakiTakahashi (approved):
LGTM!

aDaichiOta (approved):

Issue #186, opened 2021-12-07
Adding `SameSite` to `Antikythera.Http.SetCookie`


sylph01:
# Rationale

`Antikythera.Http.SetCookie` lacks the option to set the `SameSite` attribute of `Set-Cookie` header, and now it is forced into using `SameSite=Lax`.
As my team and I came across a need to set the `SameSite` directive to `SameSite=none` (especially in combination with `Antikythera.Session`), I am raising this issue and proposing the addition of this functionality. This would also help gears that want to enforce `SameSite=strict`.

# Proposed changes

- Add `same_site` field to `Antikythera.Http.SetCookie`
    - This will be an enum that takes either `:lax`, `:strict`, or `:none`
    - Adding a field under [the SetCookie module's field list](https://github.com/access-company/antikythera/blob/ac29b6439f3ab444d27d17fb30b6af0cc5a12c8d/lib/type/http.ex#L43) and adding a type would do this
    - I am ready to write up a patch for this change
- Add an interface to  `Antikythera.Plug.Session.load/2`
    - When explicitly adding a Cookie entry, passing an optional argument to `Antikythera.Conn.put_resp_cookie/4` can achieve this
    - But when used in combination with `Antikythera.Plug.Session.load/2` it is not trivial, so I would like advice on how to change this
        - As of right now, I am thinking of passing options under `:set_cookie` key, then passing this option to `make_before_send/2` (this would add an argument and thus change the signature to `make_before_send/3` ) so that it can be passed onto `Antikythera.Conn.put_resp_cookie/4` ([now called with only 3 arguments](https://github.com/access-company/antikythera/blob/47a93f3d4c70975f7296725c9bde2ea823867436/lib/web/plug/session.ex#L48)).

# Relevant references

- [SameSite cookies - HTTP | MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite)

aYukiSekiguchi:
Thank you for filing a bug!

Since [cowlib](https://ninenines.eu/docs/en/cowlib/2.9/manual/cow_cookie/ ) which Antikythera uses, has `:same_site`, `Antikythera.Http.SetCookie` should have it as well. However, cowlib 2.9 or earlier only supports `:lax` and `:strict`, so we have to update cowlib to 2.10. It means we have to update cowboy to [2.9](https://ninenines.eu/docs/en/cowboy/2.9/guide/migrating_from_2.8/#_features_added).

[...]

PR #187, opened 2021-12-08
Read Set-Cookie header options from Session plug's options

sylph01:
# Rationale

This is related to #186 (specifically "Add an interface to `Antikythera.Plug.Session.load/2`"), but it stems from a different purpose, where we needed to explicitly set the expiration period of the session cookie.
This enables the gear to specify the behavior of the session cookie via the keyword list passed to the plug.

# Part that I may need help on

- Type signature of `load/2` 's `opts` variable. The keyword list should accept a map (which is passed to `make_before_send/3`), but this specification might be too broad
    - Note: I have ran `mix dialyzer` and got no warnings
- Are there any test codes relating to this specific code?
- What to do if store is not a cookie?

### DIFF
--- a/lib/web/plug/session.ex
+++ b/lib/web/plug/session.ex
@@ -20,10 +20,14 @@ defmodule Antikythera.Plug.Session do
 
   alias Antikythera.Conn
   alias Antikythera.Session
+  alias Antikythera.Http
 
-  defun load(conn :: v[Conn.t()], opts :: Keyword.t(String.t() | atom | map)) :: Conn.t() do
+  @type load_option_t ::
+          {:key, String.t()} | {:store, atom} | {:set_cookie, Http.SetCookie.options_t()}
+
+  defun load(conn :: v[Conn.t()], opts :: v[list(load_option_t)]) :: Conn.t() do
     key             = opts[:key]
-    set_cookie_opts = opts[:set_cookie]
+    set_cookie_opts = Keyword.get(opts, :set_cookie, %{})
     store_name = Keyword.get(opts, :store, :cookie) |> Atom.to_string() |> Macro.camelize()
     store_module = Module.safe_concat("Antikythera.Session", store_name)
     {session_id, data} = store_module.load(Conn.get_req_cookie(conn, key))
--- a/lib/web/plug/session.ex
+++ b/lib/web/plug/session.ex
@@ -21,8 +21,9 @@ defmodule Antikythera.Plug.Session do
   alias Antikythera.Conn
   alias Antikythera.Session
 
-  defun load(conn :: v[Conn.t()], opts :: Keyword.t(String.t() | atom)) :: Conn.t() do
-    key = opts[:key]
+  defun load(conn :: v[Conn.t()], opts :: Keyword.t(String.t() | atom | map)) :: Conn.t() do
+    key             = opts[:key]
+    set_cookie_opts = opts[:set_cookie]
     store_name = Keyword.get(opts, :store, :cookie) |> Atom.to_string() |> Macro.camelize()
     store_module = Module.safe_concat("Antikythera.Session", store_name)
     {session_id, data} = store_module.load(Conn.get_req_cookie(conn, key))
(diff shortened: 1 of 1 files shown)


lib/web/plug/session.ex
```diff
@@ -21,8 +21,9 @@ defmodule Antikythera.Plug.Session do
   alias Antikythera.Conn
   alias Antikythera.Session
 
-  defun load(conn :: v[Conn.t()], opts :: Keyword.t(String.t() | atom)) :: Conn.t() do
-    key = opts[:key]
+  defun load(conn :: v[Conn.t()], opts :: Keyword.t(String.t() | atom | map)) :: Conn.t() do
```
irisTa56, line 24: > Type signature of `load/2` 's `opts` variable.

I think we can define types of each option and use them instead of `Keyword.t`.

```elixir
  @type load_option_t ::
          {:key, String.t()} | {:store, atom} | {:set_cookie, Http.SetCookie.options_t()}

  defun load(conn :: v[Conn.t()], opts :: v[list(load_option_t)]) :: Conn.t() do
```

lib/web/plug/session.ex
```diff
@@ -21,8 +21,9 @@ defmodule Antikythera.Plug.Session do
...
+    key             = opts[:key]
+    set_cookie_opts = opts[:set_cookie]
```
irisTa56, line 26: The default value should also be a valid `Http.SetCookie.options_t`.

```suggestion
    set_cookie_opts = Keyword.get(opts, :set_cookie, %{})
```

irisTa56:
> Are there any test codes relating to this specific code?

I couldn't find such test in this repository...
(but I found related tests [here](https://github.com/access-company/testgear/blob/master/test/session_test.exs) in `testgear` repository.)

> What to do if store is not a cookie?

I'm not sure because I don't know the detailed requirement, but
can it be solved by adding a new store module which implements `Antikythera.Session.Store`?

lib/web/plug/session.ex
sylph01: Addressed this in e534b4c

lib/web/plug/session.ex
sylph01: Also addressed this in e534b4c

lib/web/plug/session.ex
```diff
@@ -20,9 +20,14 @@ defmodule Antikythera.Plug.Session do
...
   alias Antikythera.Session
+  alias Antikythera.Http
 
-  defun load(conn :: v[Conn.t()], opts :: Keyword.t(String.t() | atom)) :: Conn.t() do
-    key = opts[:key]
+  @type load_option_t ::
+          {:key, String.t()} | {:store, atom} | {:set_cookie, Http.SetCookie.options_t()}
+
+  defun load(conn :: v[Conn.t()], opts :: v[list(load_option_t)]) :: Conn.t() do
+    key             = opts[:key]
```
aYukiSekiguchi, line 29: Could you use `mix format`?
I guess `mix format` changes this indentation.

aYukiSekiguchi (changes requested):

irisTa56 (approved):
LGTM!

Fumipo-Theta (approved):
LGTM!

aYukiSekiguchi:
> > Are there any test codes relating to this specific code?
> 
> I couldn't find such test in this repository...
> but I found related tests here in testgear repository.

I don't think you need to write test code for testgear, but I want to confirm that you manually tested this.

aYukiSekiguchi:
> What to do if store is not a cookie?

I'm not sure what `Uses cookie store by default` means. However, it looks like session data is always stored in cookie, and `Antikythera.Session.Cookie` just encodes/decodes data to/from JSON. Therefore, your `set_cookie_opts` looks correct approach.

aYukiSekiguchi (approved):
LGTM if you have tested this PR

sylph01:
please wait while I prepare some evidence :bow:

aMasakiTakahashi (approved):
LGTM!

sylph01:
I got some evidence of how this works.

## Prerequisites

- `antikythera_dep = {:antikythera, [github: "sylph01/antikythera", ref: "2479a8d8cad17c63d000782b5c0e432ade970aca"]}` in `antikythera_instance_example` 's `mix.exs`
    - the commit hash points to the HEAD of this branch
- then in local testgear, do following to compile/run with specified version of antikythera (change path to your environment accordingly):

```
ANTIKYTHERA_INSTANCE_DEP='{:antikythera_instance_example, [git: "/home/sylph01/work/antikythera_instance_example"]}' mix deps.get
ANTIKYTHERA_INSTANCE_DEP='{:antikythera_instance_example, [git: "/home/sylph01/work/antikythera_instance_example"]}' mix compile
ANTIKYTHERA_INSTANCE_DEP='{:antikythera_instance_example, [git: "/home/sylph01/work/antikythera_instance_example"]}' iex -S mix
```

## Actual evidence

In testgear's `/web/controller/session.ex`, I changed the plug line to add `max_age` option:

Screenshot_20220121_153507

### Before

Screenshot_20220121_153134

### After

Screenshot_20220121_153305

Note that the new Set-Cookie header has the Max-Age value.

aYukiSekiguchi:
Great! Thank you!

PR #188, opened 2021-12-13
[238866] Temporally add log for Config.Gear.read/1 error

aYukiSekiguchi:
https://acsmine.tok.access-company.com/redmine/issues/238866

### DIFF
--- a/core/config/gear.ex
+++ b/core/config/gear.ex
@@ -49,6 +49,14 @@ defmodule AntikytheraCore.Config.Gear do
           {:error, reason} ->
             msg = "failed to decode gear config JSON (#{gear_name}): #{inspect(reason)}"
             L.error(msg)
+            L.error("Raw data for (#{gear_name}: #{inspect(encrypted, limit: :infinity)}")
+
+            L.error(
+              "Decrypted data for (#{gear_name}: #{
+                inspect(Aes.ctr128_decrypt(encrypted, EncryptionKey.get()))
+              }"
+            )
+
             raise msg
         end
 
--- a/core/config/gear.ex
+++ b/core/config/gear.ex
@@ -49,6 +49,7 @@ defmodule AntikytheraCore.Config.Gear do
           {:error, reason} ->
             msg = "failed to decode gear config JSON (#{gear_name}): #{inspect(reason)}"
 
+            # TODO: this log should be removed after fixing error
             L.error(
               Enum.join(
                 [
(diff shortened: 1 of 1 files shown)


Fumipo-Theta (approved):
LGTM!

aMasakiTakahashi (approved):
LGTM!

core/config/gear.ex
```diff
@@ -49,6 +49,14 @@ defmodule AntikytheraCore.Config.Gear do
           {:error, reason} ->
             msg = "failed to decode gear config JSON (#{gear_name}): #{inspect(reason)}"
             L.error(msg)
+            L.error("Raw data for (#{gear_name}: #{inspect(encrypted, limit: :infinity)}")
```
irisTa56, line 52: If this is NOT temporary code for investigation, I think

- We don't need to call `Aes.ctr128_decrypt/2` again.
- Putting all information into one `L.error/1` might be better because they are all related.

So I'd like to suggest the following code:

```elixir
        decrypt_result = Aes.ctr128_decrypt(encrypted, EncryptionKey.get())

        case R.bind(decrypt_result, &decode/1) do
          {:ok, conf} ->
            conf

          {:error, reason} ->
            msg = "failed to decode gear config JSON (#{gear_name}): #{inspect(reason)}"

            L.error(
              Enum.join(
                [
                  msg,
                  "Raw data: " <> inspect(encrypted, limit: :infinity),
                  "Decrypted data: " <> inspect(decrypt_result)
                ],
                "\n"
              )
            )

            raise msg
        end
```

aYukiSekiguchi: I thought it should be temporal, so I changed the title of this PR.

Then, I don't want to change logic for temporal implementation. I'd like to keep the logic for `R.bind/2`.

Your error log format looks good! Fixed at https://github.com/access-company/antikythera/pull/188/commits/cd57fd22b0db8d2a5cd3ad1b5dba63f7679abdae

irisTa56: Thank you for the fix!

In addition, could you add a TODO comment like `# TODO: this log should be removed after fixing error` for reminder?

irisTa56 (approved):
I left only one minor comment.
LGTM!

core/config/gear.ex
aYukiSekiguchi: Yeah, good suggestion! Fixed at https://github.com/access-company/antikythera/pull/188/commits/542d5d637cf42373299654a7ce42df817d0bb30e

PR #191, opened 2021-12-16
[238866] Atomically write conf and settings using rename

aYukiSekiguchi:
https://acsmine.tok.access-company.com/redmine/issues/238866

### DIFF
--- a/core/config/core.ex
+++ b/core/config/core.ex
@@ -66,7 +66,14 @@ defmodule AntikytheraCore.Config.Core do
   defun write(config :: v[map]) :: :ok do
     path = CorePath.core_config_file_path()
     content = Aes.ctr128_encrypt(inspect(config), EncryptionKey.get())
-    File.write!(path, content, [:sync])
+    node_name = Node.self() |> Atom.to_string()
+    # Convert `#PID<0.23069.636>` to `0.23069.636`
+    pid_path_safe_string =
+      inspect(self()) |> String.split("<") |> Enum.at(1) |> String.replace(">", "")
+
+    temp_path = "#{path}-#{node_name}-#{pid_path_safe_string}"
+    File.write!(temp_path, content, [:sync])
+    File.rename!(temp_path, path)
   end
 
   defun dump_from_env_to_file() :: :ok do
--- a/core/config/core.ex
+++ b/core/config/core.ex
@@ -66,14 +66,7 @@ defmodule AntikytheraCore.Config.Core do
   defun write(config :: v[map]) :: :ok do
     path = CorePath.core_config_file_path()
     content = Aes.ctr128_encrypt(inspect(config), EncryptionKey.get())
-    node_name = Node.self() |> Atom.to_string()
-    # Convert `#PID<0.23069.636>` to `0.23069.636`
-    pid_path_safe_string =
-      inspect(self()) |> String.split("<") |> Enum.at(1) |> String.replace(">", "")
-
-    temp_path = "#{path}-#{node_name}-#{pid_path_safe_string}"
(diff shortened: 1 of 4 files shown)


core/config/core.ex
```diff
@@ -66,7 +66,14 @@ defmodule AntikytheraCore.Config.Core do
...
     content = Aes.ctr128_encrypt(inspect(config), EncryptionKey.get())
-    File.write!(path, content, [:sync])
+    node_name = Node.self() |> Atom.to_string()
+    # Convert `#PID<0.23069.636>` to `0.23069.636`
+    pid_path_safe_string =
+      inspect(self()) |> String.split("<") |> Enum.at(1) |> String.replace(">", "")
+
+    temp_path = "#{path}-#{node_name}-#{pid_path_safe_string}"
+    File.write!(temp_path, content, [:sync])
+    File.rename!(temp_path, path)
```
irisTa56, line 76: I have three suggestions:

- Using `AntikytheraCore.Cluster.NodeId.get/0` instead of `Node.self/0`
    - I think it's more simple and light.
- Adding a common function to [`core/util/path.ex`](https://github.com/access-company/antikythera/blob/590391d664de542b28fd3eb51ca37ac1242c1775/core/util/path.ex#L121) and using it here
    - I think the same logic is used for core/gear config and tenant setting.
    - As the common function, I imagine something like the following:

      ```elixir
      def write_atomic(path, content) do
          node_name = ...
          pid_path_safe_string = ...
          temp_path = "#{path}-#{node_name}-#{pid_path_safe_string}"
          File.write!(temp_path, content, [:sync])
          File.rename!(temp_path, path)
      end
      ```

- Suffixing a temporary file name with `-tmp` and rejecting such file [here](https://github.com/access-company/antikythera/blob/590391d664de542b28fd3eb51ca37ac1242c1775/core/util/path.ex#L139-L140)
    - If the thetemporary file remains (e.g. due to failure of write/rename operation), it will be [detected and a new atom will be generated](https://github.com/access-company/antikythera/blob/590391d664de542b28fd3eb51ca37ac1242c1775/core/config/gear.ex#L93-L95).
    - (Only checking `String.contains?("-")` might be ok because [gear and tenant names don't have `-`](https://github.com/access-company/antikythera/blob/75e82e4b6ef75950c592a1ee154778d544ae9133/lib/type/string.ex#L184-L193).)

aYukiSekiguchi: Good catch!

I thought this function is not so generic because it assume that one Erlang process doesn't use this function in parallel. 
This was the reason why I didn't create it in util/. However, I remembered that there is no thread in an Erlang process, so the assumption is generic enough.

I'm not sure `write_atomic` is common words in Elixir, but I think `atomic_xxx` is common in programming world like [these C++ methods](https://en.cppreference.com/w/cpp/atomic/atomic_load) or [MariaDB](https://mariadb.com/kb/en/atomic-write-support/). I changed your function name.

Fixed at https://github.com/access-company/antikythera/pull/191/commits/670178cc4987727143dcc80e60f9989ab85c9ab5

irisTa56: > I thought this function was not so generic because it assumes that one Erlang process doesn't use this function in parallel.
> This was the reason why I didn't create it in util/. However, I remembered that there is no thread in an Erlang process, so the assumption is generic enough.

Thank you for explanation!
I didn't think so deeply but now I got your intention.

For the function name, my suggestion was just idea and `atomic_write!/2` looks nice to me!

irisTa56 (approved):
I left only an optional comment. LGTM!

Fumipo-Theta (approved):
LGTM!

aMasakiTakahashi (approved):
LGTM!

PR #199, opened 2022-03-16
[243759] Do not delete auto generated module in appup

irisTa56:
https://acsmine.tok.access-company.com/redmine/issues/243759#note-4

### DIFF
--- a/core/mix/appup.ex
+++ b/core/mix/appup.ex
@@ -53,6 +53,10 @@ defmodule AntikytheraCore.Release.Appup do
     {only_v1, only_v2, diff_pairs} = :beam_lib.cmp_dirs(v1_ebin_dir, v2_ebin_dir)
     diff = reject_unchanged_modules(diff_pairs) |> Enum.map(fn {_f1, f2} -> f2 end)
 
+    # Croma automatically defines a module which might be common across multiple gears,
+    # e.g. `Elixir.Croma.TypeGen.Nilable.Antikythera.Url`.
+    # Such module must not be deleted when upgrading one of the gears
+    # because the other gears are still using that module.
     only_v1_without_auto_generated =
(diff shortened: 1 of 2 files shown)


aMasakiTakahashi (approved):
LGTM!

core/mix/appup.ex
```diff
@@ -53,13 +53,24 @@ defmodule AntikytheraCore.Release.Appup do
...
+
     file_content = {
       v2_charlist,
       [
-        {v1_charlist, generate_instructions(only_v2, diff, only_v1)}
+        {v1_charlist, generate_instructions(only_v2, diff, only_v1_without_auto_generated)}
       ],
       [
-        {v1_charlist, generate_instructions(only_v1, diff, only_v2)}
+        {v1_charlist, generate_instructions(only_v1_without_auto_generated, diff, only_v2)}
```
aYukiSekiguchi, line 73: I understand this isn't used because Antikythera won't use downgrade. However, if this is used, this may cause `undefined module error`. e.g. a new Antikythera instance which doesn't load `Elixir.Croma.TypeGen.Nilable.Antikythera.Url` downgrades a gear which needs `Elixir.Croma.TypeGen.Nilable.Antikythera.Url`.

If downgrading is used, this `only_v2` may have the same problem. I think it's no problem. However, I don't think we should implement a new bug for downgrading even if it isn't used.

```suggestion
        {v1_charlist, generate_instructions(only_v1, diff, only_v2)}
```

irisTa56: At first I prefer symmetry.
But, from practical point of view, I understand that your suggestion is better to prevent future problem.
So I committed the suggestion.
Thank you!

aYukiSekiguchi (approved):
LGTM

Fumipo-Theta (approved):
LGTM!

PR #200, opened 2022-03-24
[243759] Do not use delete_module instruction in appup

irisTa56:
https://acsmine.tok.access-company.com/redmine/issues/243759#note-11

### DIFF
--- a/core/mix/appup.ex
+++ b/core/mix/appup.ex
@@ -56,15 +56,10 @@ defmodule AntikytheraCore.Release.Appup do
     file_content = {
       v2_charlist,
       [
-        # Since some modules might be common across multiple gears,
-        # delete_module instructions might accidentally delete modules in use.
-        # So we don't delete modules in the upgrade instruction.
-        # Note that, as a consequence of the above,
-        # we need to deploy without hot code upgrade to delete really unused modules.
-        {v1_charlist, generate_instructions(only_v2, diff, [])}
+        {v1_charlist, generate_instructions(only_v2, diff)}
       ],
       [
-        {v1_charlist, generate_instructions(only_v1, diff, only_v2)}
+        {v1_charlist, generate_instructions(only_v1, diff)}
       ]
     }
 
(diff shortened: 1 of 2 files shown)


core/mix/appup.ex
```diff
@@ -53,21 +53,15 @@ defmodule AntikytheraCore.Release.Appup do
...
     file_content = {
       v2_charlist,
       [
-        {v1_charlist, generate_instructions(only_v2, diff, only_v1_without_auto_generated)}
+        # Since some modules might be common across multiple gears,
+        # delete_module instructions might accidentally delete modules in use.
+        # So we don't delete modules in the upgrade instruction.
+        # Note that, as a consequence of the above,
+        # we need to deploy without hot code upgrade to delete really unused modules.
+        {v1_charlist, generate_instructions(only_v2, diff, [])}
```
aYukiSekiguchi, line 64: For the same reason in https://github.com/access-company/antikythera/pull/199#discussion_r827684985,
```elixir
{v1_charlist, generate_instructions(only_v1, diff, only_v2)}
```
in the next block should be 
```elixir
{v1_charlist, generate_instructions(only_v1, diff, [])}
```

If both of `gear1` and `gear2` has `module M` and `gear1` depends on `gear2`,  IIUC, the app file for `gear1` doesn't have `module M`. Then, if `gear1` removes the dependency for `gear2`, the appup file for the new `gear1` have `:add_module, M` in upgrade, and `:delete_module, M` in downgrade. The downgrade instruction deletes `module M` which both of `gear1` and `gear2` depends on.

core/mix/appup.ex
```diff
@@ -53,21 +53,15 @@ defmodule AntikytheraCore.Release.Appup do
...
-        beam_file
-        |> read_module_name()
-        |> AntikytheraCore.Version.Gear.auto_generated_module?()
-      end)
-
     file_content = {
       v2_charlist,
       [
-        {v1_charlist, generate_instructions(only_v2, diff, only_v1_without_auto_generated)}
+        # Since some modules might be common across multiple gears,
```
aYukiSekiguchi, line 59: I think `some modules` should be `some generated modules` or `some modules which are generated by macro` or something is better. We assume that an Antikythera instance has common modules, i.g. libraries, in its `mix.exs`, so there is no problem in most cases. The exception is modules which are generated by macro.

core/mix/appup.ex
irisTa56: Thank you for your comment.
Sorry for not considering downgrade case again.
(Maybe the fix in my previous PR was not enough too...)

I eliminated `:delete_module` by redefining `generate_instructions/2`.
https://github.com/access-company/antikythera/pull/200/commits/b7059e1d9b21b07341c3d4c6fc70311bb958c3d2

core/mix/appup.ex
irisTa56: Thank you.
It's important to make our concerns as clear as possible.
I totally revised the comment.
https://github.com/access-company/antikythera/pull/200/commits/b7059e1d9b21b07341c3d4c6fc70311bb958c3d2

aYukiSekiguchi (approved):
LGTM

aMasakiTakahashi (approved):
LGTM!

Fumipo-Theta (approved):
LGTM!

PR #207, opened 2022-10-13
[239880] Make ensure_antikythera_instance_compiled/0 more compatible

Double-oxygeN:
https://acsmine.tok.access-company.com/redmine/issues/239880

### DIFF
--- a/core/mix/generate_release.ex
+++ b/core/mix/generate_release.ex
@@ -73,10 +73,16 @@ defmodule Mix.Tasks.AntikytheraCore.GenerateRelease do
     end
   end
 
+  @dialyzer {:no_match, ensure_antikythera_instance_compiled: 0}
   defp ensure_antikythera_instance_compiled() do
     try do
       # Generate <antikythera_instance>.app as it can be for older version of the instance.
-      :ok = Mix.Tasks.Compile.App.run(["--force"])
+      if System.version() |> Version.parse!() |> Version.match?("~> 1.10") do
+        {:ok, _} = Mix.Tasks.Compile.App.run(["--force"])
+      else
+        # TODO: remove this pattern and the suppression just above after upgrading to Elixir 1.10+
+        :ok = Mix.Tasks.Compile.App.run(["--force"])
+      end
     rescue
       File.Error ->
         # Directory for <antikythera_instance>.app does not exist,


core/mix/generate_release.ex
```diff
@@ -73,10 +73,16 @@ defmodule Mix.Tasks.AntikytheraCore.GenerateRelease do
     end
   end
 
+  @dialyzer {:no_match, ensure_antikythera_instance_compiled: 0}
```
irisTa56, line 76: This workaround might be unnecessary if we determine Elixir version at compile time.
My suggestion is to define a function (like the following) based on Elixir version, and call it in `ensure_antikythera_instance_compiled/0`.

```elixir
  if Version.match?(System.version(), "~> 1.10") do
    defp force_compile_app(expr) do
      {:ok, _} = Mix.Tasks.Compile.App.run(["--force"])
    end
  else
    defp force_compile_app(expr) do
      :ok = Mix.Tasks.Compile.App.run(["--force"])
    end
  end
```

By doing so, I think that only the correct pattern goes into compiled code (`{:ok, _} =` for 1.10+, `:ok =` for other versions).

core/mix/generate_release.ex
```diff
@@ -73,10 +73,16 @@ defmodule Mix.Tasks.AntikytheraCore.GenerateRelease do
...
   defp ensure_antikythera_instance_compiled() do
     try do
       # Generate <antikythera_instance>.app as it can be for older version of the instance.
-      :ok = Mix.Tasks.Compile.App.run(["--force"])
+      if System.version() |> Version.parse!() |> Version.match?("~> 1.10") do
```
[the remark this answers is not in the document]
Double-oxygeN: Fixed at c1903af

core/mix/generate_release.ex
Double-oxygeN: I appreciate your code suggestion.
It works, but dialyzer gives me an error about `:no_return` in Elixir 1.10.4.
I am investigating the cause.

aYukiSekiguchi: I think `force_compile_app!/1` is better name because this function may throw `MatchError`.

Double-oxygeN: I tried to avoid the `:no_return` warning, but I could not find the right way in Elixir 1.10.4.
Probably dialyzer misinterprets `Mix.Tasks.Compile.App.run/1` as a function that always fails.

The following code passes the dialyzer check even in Elixir 1.10.4, but I don't believe that this solution is the best.

```elixir
  if Version.match?(System.version(), "~> 1.10") do
    if Version.match?(System.version(), "~> 1.10.0"), do: @spec force_compile_app!() :: no_return()
    defp force_compile_app!() do
      {:ok, _} = Mix.Tasks.Compile.App.run(["--force"])
    end
  else
    defp force_compile_app!() do
      :ok = Mix.Tasks.Compile.App.run(["--force"])
    end
  end
```

irisTa56: Wrong spec is a cause of the `:no_return` warning.
(I think dialyser itself causes much less false positive than false negative.)

- `Mix.Tasks.Compile.App.run/1` calls `load_compile_env/1` at [this line](https://github.com/elixir-lang/elixir/blob/v1.10.4/lib/mix/lib/mix/tasks/compile.app.ex#L146)
- `load_compile_env/1` calls `Mix.ProjectStack.compile_env/1` with `:unset` as its argument
- But `Mix.ProjectStack.compile_env/1`'s [spec](https://github.com/elixir-lang/elixir/blob/v1.10.4/lib/mix/lib/mix/project_stack.ex#L108) is `compile_env([term]) :: [term]`, which is inconsistent
    - (You can find this inconsistent by running `mix dialyzer` against [mix itself](https://github.com/elixir-lang/elixir/tree/main/lib/mix). Note that it's required to add `dialyxir` as dependency.)

This problem has been fixed by [this commit](https://github.com/elixir-lang/elixir/commit/86534cf7ec8d633622f58f27cefd77604b335583), and seems to have had no effect on runtime behavior.

Since [dialyxir's document](https://github.com/jeremyjh/dialyxir/blob/1.2.0/README.md?plain=1#L264) says

> A better option is to ignore the specific warnings you can't fix (maybe due to a bug upstream, ...

I think it's better to add the following lines near other module attributes.

```elixir
  # TODO: Remove after requiring Elixir 1.11+
  if Version.match?(System.version(), "~> 1.10.0") do
    @dialyzer {:no_return, force_compile_app!: 0}
  end
```

Double-oxygeN: Thank you for your help!
This helps me clearly know why the function leads `:no_return` warning.
Fixed at 55f6adf. It is OK for both Elixir 1.9.4 and 1.10.4 (and so is for 1.11.4).

core/mix/generate_release.ex
```diff
@@ -85,6 +89,16 @@ defmodule Mix.Tasks.AntikytheraCore.GenerateRelease do
     end
   end
 
+  if Version.match?(System.version(), "~> 1.10") do
```
irisTa56, line 92: In order not to forget, TODO comment might be helpful.

```suggestion
  # TODO: Remove this conditional branching after requiring Elixir 1.10+
  if Version.match?(System.version(), "~> 1.10") do
```

irisTa56 (approved):
LGTM!

core/mix/generate_release.ex
Double-oxygeN: Added at 4547472.

aMasakiTakahashi (approved):
LGTM!

aYukiSekiguchi (approved):
LGTM

PR #208, opened 2022-10-21
[239881] Prepare for migration to Elixir 1.11.4

Double-oxygeN:
https://acsmine.tok.access-company.com/redmine/issues/239881#note-21

### DIFF
--- a/core/core.ex
+++ b/core/core.ex
@@ -16,6 +16,7 @@ defmodule AntikytheraCore do
   """
   @impl true
   def start(_type, _args) do
+    Logger.add_translator({AntikytheraCore.ErlangLogTranslator, :translate})
     L.info("starting AntikytheraCore")
     add_gears_dir_to_erl_libs()
     AntikytheraCore.FileSetup.setup_files_and_ets_tables()
--- a/core/core.ex
+++ b/core/core.ex
@@ -17,6 +17,7 @@ defmodule AntikytheraCore do
   @impl true
   def start(_type, _args) do
     Logger.add_translator({AntikytheraCore.ErlangLogTranslator, :translate})
+    Logger.configure(level: :info)
     L.info("starting AntikytheraCore")
     add_gears_dir_to_erl_libs()
     AntikytheraCore.FileSetup.setup_files_and_ets_tables()
--- a/core/core.ex
+++ b/core/core.ex
@@ -17,6 +17,9 @@ defmodule AntikytheraCore do
   @impl true
   def start(_type, _args) do
     Logger.add_translator({AntikytheraCore.ErlangLogTranslator, :translate})
+    # In dev or local environment, the log level is initially set to `:notice` at mix_common.exs
+    # in order to avoid SASL progress reports.
+    # The log level is restored to `:info` after loading antikythera.
     Logger.configure(level: :info)
     L.info("starting AntikytheraCore")
     add_gears_dir_to_erl_libs()
--- a/mix_common.exs
+++ b/mix_common.exs
@@ -158,6 +158,9 @@ defmodule Antikythera.MixConfig do
   end
 
   defp default_configs() do
+    # Loading log translator
+    Code.eval_file("erlang_log_translator.ex", Path.join(__DIR__, "core/util"))
+
     [
       # Logger configurations.
       sasl: [
--- a/mix_common.exs
+++ b/mix_common.exs
@@ -158,6 +158,20 @@ defmodule Antikythera.MixConfig do
   end
 
   defp default_configs() do
+    # TODO: Remove after requiring Elixir 1.11+
+    logger_configs_for_compatibility =
+      if Version.match?(System.version(), "~> 1.11") do
+        [
+          # To suppress progress reports during start-up in development environments
+          level: (if Mix.env() == :prod, do: :info, else: :notice)
+        ]
+      else
+        [
+          level: :info,
+          translators: [{AntikytheraCore.ErlangLogTranslator, :translate}]
+        ]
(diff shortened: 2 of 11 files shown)


mix_common.exs
```diff
@@ -94,11 +94,12 @@ defmodule Antikythera.MixCommon do
     case Mix.env() do
       :dev -> [
         :mix,     # Suppress warning about Mix.Task behaviour
-        :eex,     # Only used during compilation, suppress warning about EEx.Engine behaviour
         :ex_unit, # Suppress warnings about calling ExUnit functions in Antikythera.Test.*
       ]
       _ -> []
-    end
+    end ++ [
+      :eex,       # Only used during compilation, suppress warning about EEx.Engine behaviour
```
aYukiSekiguchi, line 101: Since EEx is used by macro, a gear shouldn't depend on `:eex` at runtime.
After your fix, a gear depends on `:eex` at runtime.
> _build/prod//lib/gear_name/ebin/gear_name.app:                             ex_aws_ses,sweet_xml,dodai_client_elixir,eex]},

I think the 3rd option is better like `suppress warning about EEx.Engine behaviour` comment here.
>   3. In case you don't want to add a requirement to :eex, you may optionally skip this warning by adding [xref: [exclude: [EEx]]] to your "def project" in mix.exs

mix_common.exs
```diff
@@ -157,13 +158,10 @@ defmodule Antikythera.MixConfig do
...
 
   defp default_configs() do
-    [
-      # Limit port range to be used for ErlangVM-to-ErlangVM communications.
-      kernel: [
-        inet_dist_listen_min: 6000,
-        inet_dist_listen_max: 7999,
-      ],
+    # Loading log translator
+    Code.eval_file("erlang_log_translator.ex", Path.join(__DIR__, "core/util"))
```
aYukiSekiguchi, line 162: There is no `:translators` in [the config doc](https://hexdocs.pm/logger/1.12.3/Logger.html#module-configuration), and [the Translator module doc](https://hexdocs.pm/logger/1.12.3/Logger.Translator.html#content) mentions only `Logger.add_translator/1`. Therefore, I guess `:translators` isn't an API. This is the reason why warning is printed.

Since `AntikytheraCore.ErlangLogTranslator` should be used at runtime instead of compile time, I think it would be better to add `AntikytheraCore.ErlangLogTranslator` at `AntikytheraCore.start/2`.

aYukiSekiguchi (changes requested):

mix_common.exs
Double-oxygeN: Thank you for your suggestion.
Fixed at 263be31 and 3a2ea42.
This change prevents to skip some of the logs before `AntikytheraCore` starts.

irisTa56: > This change prevents to skip some of the logs before `AntikytheraCore` starts.

I found that many (200~300 lines) noisy logs show in terminal during gear's `mix test` and `iex -S mix`.
I think gear developers will have difficulty navigating the terminal.
Could you fix this problem?
(Sorry, I currently don't have any suggestions.)

Double-oxygeN: 04463a7 may be the simplest solution.

irisTa56: I think it's better to avoid setting `:translators` if not needed (because it's undocumented).
How about defining the version-dependent part separately and appending it to `:logger` configs?

```elixir
# TODO: Remove after requiring Elixir 1.11+
logger_configs_for_compatibility =
  if Version.match?(System.version(), "~> 1.11") do
    [
      # To suppress progress reports during start-up in development environments
      level: (if Mix.env() == :prod, do: :info, else: :notice)
    ]
  else
    [
      level: :info,
      translators: [{AntikytheraCore.ErlangLogTranslator, :translate}]
    ]
  end
```

Double-oxygeN: It looks good.
Fixed at d8a5606.

core/core.ex
```diff
@@ -16,6 +16,8 @@ defmodule AntikytheraCore do
   """
   @impl true
   def start(_type, _args) do
+    Logger.add_translator({AntikytheraCore.ErlangLogTranslator, :translate})
+    Logger.configure(level: :info)
```
aYukiSekiguchi, line 20: It would be better to write a comment about why this setting is needed

Double-oxygeN:
```elixir
Logger.add_translator({AntikytheraCore.ErlangLogTranslator, :translate})
# In dev or local environment, the log level is initially set to notice
# in order to avoid SASL progress reports.
# The log level is integrated into info after loading antikythera.
Logger.configure(level: :info)
```

How about this? Could you please tell me if you have any other ideas or better phrases.

aYukiSekiguchi: Looks good! But I'd like to change slightly.

```elixir
# In dev or local environment, the log level is initially set to `:notice` at mix_common.exs
# in order to avoid SASL progress reports.
# The log level is restored to `:info` after loading antikythera.
```

* Add `at mix_common.exs` to notify what file changed to `:notice`
    * Someone may want to see info level log during start up process to debug and find this `Logger.configure`.
* Change `integrated` to `restored`
    * IMHO, `integrated` is more difficult word than `restored`
* Use Elixir atoms, e.g. `:notice`, instead of English words

Double-oxygeN: Thank you for your advice!
Added at ab9bf44.

aYukiSekiguchi (approved):
LGTM

irisTa56 (approved):
LGTM!

aMasakiTakahashi (approved):
LGTM!

PR #209, opened 2022-11-10
[253873] Update hex libraries

Double-oxygeN:
https://acsmine.tok.access-company.com/redmine/issues/253873#note-15

PR #210, opened 2022-12-07
Revert "[253873] Update hex libraries (#209)"

Double-oxygeN:
https://acsmine.tok.access-company.com/redmine/issues/255215#note-11

This reverts commit 6db73defe975369ec44897d1705befc3e5952020 (#209).
This is necessary to add URL empty path detection **before** updating Hackney.


irisTa56 (approved):
LGTM!

aYukiSekiguchi (approved):
LGTM

> This is necessary to add URL empty path detection before updating Hackney.

It's hard to understand this description. This description focuses on what you're doing now. You should write why this revert is needed.

I think you should describe about the unintended API change due to the hackney update. Then, you'll add logging in another PR and update hackney with the API change in another PR.

Double-oxygeN:
> It's hard to understand this description. This description focuses on what you're doing now. You should write why this revert is needed.
> 
> I think you should describe about the unintended API change due to the hackney update. Then, you'll add logging in another PR and update hackney with the API change in another PR.

I apologize for the lack of explanation.

The reason why this revert must be to do is because Hackney, which is updated by #209, has a problem about handling URLs with empty path while sending HTTP requests.
Hackney 0.16.0 or later sends illegal HTTP requests for URLs with empty path, such as `https://example.com?q=query`.
Therefore, the function for detecting such URLs (#211) is needed for debugging running gears before updating the library.

* * *

By the way, thank you all for reviewing this PR.

PR #213, opened 2022-12-26
[239881] Disable protocol consolidation

Double-oxygeN:
https://acsmine.tok.access-company.com/redmine/issues/239881#note-78

* * *

## Purpose

* To avoid inclusion of consolidated protocol information in the core PLT file in Elixir 1.11+.

## Reason

* In Elixir 1.11+, dialyxir becomes to make a core PLT file with project-specific consolidated protocol information.
  * Maybe [this commit](/elixir-lang/elixir/commit/4bd3c9936ba4d6d72b662655bfd4ab7c5a283e0a) caused it.
* In other words, the core PLT file is rewritten for each type checking.
  * It fails if the operator (e.g. CI/CD tool) does not have write permission for the file.
* Disabling protocol consolidation resolves this issue.

## Effect

* There seems  to be no performance degradation in release.
  * Release builds of Antikythera have not used protocol consolidation for a long time.

### DIFF
--- a/mix_common.exs
+++ b/mix_common.exs
@@ -11,13 +11,14 @@
 defmodule Antikythera.MixCommon do
   def common_project_settings() do
     [
-      elixir:            "~> 1.9",
-      elixirc_options:   [warnings_as_errors: true],
-      build_path:        build_path(),
-      build_embedded:    Mix.env() == :prod,
-      test_coverage:     [tool: ExCoveralls],
-      preferred_cli_env: [coveralls: :test, "coveralls.detail": :test, "coveralls.html": :test, "antikythera_local.upgrade_compatibility_test": :test],
-      xref:              [exclude: [EEx, EEx.Engine]], # Suppress undefined application warnings
+      elixir:                "~> 1.9",
+      elixirc_options:       [warnings_as_errors: true],
+      build_path:            build_path(),
+      build_embedded:        Mix.env() == :prod,
+      test_coverage:         [tool: ExCoveralls],
+      preferred_cli_env:     [coveralls: :test, "coveralls.detail": :test, "coveralls.html": :test, "antikythera_local.upgrade_compatibility_test": :test],
+      xref:                  [exclude: [EEx, EEx.Engine]], # Suppress undefined application warnings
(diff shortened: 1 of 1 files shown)


mix_common.exs
```diff
@@ -11,13 +11,14 @@
...
-      preferred_cli_env: [coveralls: :test, "coveralls.detail": :test, "coveralls.html": :test, "antikythera_local.upgrade_compatibility_test": :test],
-      xref:              [exclude: [EEx, EEx.Engine]], # Suppress undefined application warnings
+      elixir:                "~> 1.9",
+      elixirc_options:       [warnings_as_errors: true],
+      build_path:            build_path(),
+      build_embedded:        Mix.env() == :prod,
+      test_coverage:         [tool: ExCoveralls],
+      preferred_cli_env:     [coveralls: :test, "coveralls.detail": :test, "coveralls.html": :test, "antikythera_local.upgrade_compatibility_test": :test],
+      xref:                  [exclude: [EEx, EEx.Engine]], # Suppress undefined application warnings
+      consolidate_protocols: false, # Avoid inclusion of consolidated protocol information in the core PLT file
```
aYukiSekiguchi, line 21: You should write a comment that our release build doesn't use consolidate protocols.

If I look at this comment without prior knowledge, you disable consolidate protocols because you don't want to include them to the core PLT file although the disabling degrades performance in release. However, you know our release doesn't use consolidate protocols, so there is no performance degradation in release.

aYukiSekiguchi:
Could you write few line simple description in the description? ACCESS members can see the Redmine ticket, but OSS members can't see it. It would be better if PRs have simple description for OSS members.

Double-oxygeN:
> Could you write few line simple description? ACCESS members can see the Redmine ticket, but OSS members can't see it. It would be better if PRs have simple description for OSS members.

I edited the description.

mix_common.exs
Double-oxygeN: Wrote the comment at 48f1713.

aYukiSekiguchi (approved):
LGTM

aMasakiTakahashi (approved):
LGTM!

irisTa56 (approved):
LGTM!

PR #230, opened 2023-06-12
[260138] Add a test case for the error value in Httpc.post/4

aYukiSekiguchi:
https://acsmine.tok.access-company.com/redmine/issues/260138


Double-oxygeN (approved):
LGTM

PR #232, opened 2023-06-26
[239901] Use `Config` instead of deprecated `Mix.Config`

Double-oxygeN:
https://acsmine.tok.access-company.com/redmine/issues/239901


aYukiSekiguchi (approved):
LGTM

I checked the implementation and confirmed that `config/2`s, `eval!/2` and `read!/2` and `persist/1` and `put_all_env/2` are (mostly) identical.

PR #235, opened 2023-07-07
[262976] Add per-API timeout

Double-oxygeN:
https://acsmine.tok.access-company.com/redmine/issues/262976

### DIFF
--- a/lib/web/router/impl.ex
+++ b/lib/web/router/impl.ex
@@ -7,9 +7,9 @@ defmodule Antikythera.Router.Impl do
   Internal functions to implement request routing.
   """
 
-  alias Antikythera.{Http.Method, PathSegment, PathInfo, Request.PathMatches}
+  alias Antikythera.{Http.Method, PathSegment, PathInfo, Request.PathMatches, GearActionTimeout}
 
-  @typep route_entry :: {Method.t(), String.t(), module, atom, Keyword.t(String.t())}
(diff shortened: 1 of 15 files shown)


lib/web/router/impl.ex
```diff
@@ -232,10 +243,11 @@ defmodule Antikythera.Router.Impl do
           controller :: v[module],
           action :: v[atom],
           path_matches :: Keyword.t(String.t()),
-          websocket? :: v[boolean]
+          websocket? :: v[boolean],
+          timeout :: v[GearActionTimeout.t()] \\ GearActionTimeout.default()
         ) :: route_result do
     if Enum.all?(path_matches, fn {_placeholder, match} -> String.printable?(match) end) do
-      {controller, action, Map.new(path_matches), websocket?}
+      {controller, action, Map.new(path_matches), websocket?, timeout}
```
aYukiSekiguchi, line 250: I think we should check `GearActionTimeout.max/0` at runtime as well.
For example, if a gear was compiled with 5000ms, then we change the setting in the antikythera instance to 3000ms. I think we want to change the timeout of the gear to 3000ms without recompiling the gear.

Double-oxygeN: Thank you for your review.
As you said, it is necessary to check if the timeout is in range also at runtime.
Fixed at e073767.

aYukiSekiguchi (approved):
LGTM

PR #247, opened 2023-11-27
[284179] Use npm audit --audit-level critical

aYukiSekiguchi:
https://acsmine.tok.access-company.com/redmine/issues/284179


Double-oxygeN (commented):
The `--parseable` option no longer has any effects on the result. It can be omitted.

How about `--audit-level=critical`? It will notice critical dependencies via status code.
https://docs.npmjs.com/cli/v8/commands/npm-audit#audit-level

https://github.com/access-company/antikythera/blob/ff79b5b832c0406401505032e141b2803784b71b/lib/mix/prepare_assets.ex#L138

PR #253, opened 2023-12-20
[284940] Add an option which specifies :docs to GearProject

aYukiSekiguchi:
https://acsmine.tok.access-company.com/redmine/issues/284940


Double-oxygeN (approved):
LGTM

PR #260, opened 2024-03-12
[287754] HTTP client partially accepts hackney's :connect_options

irisTa56:
https://acsmine.tok.access-company.com/redmine/issues/287754

### DIFF
--- a/lib/util/httpc.ex
+++ b/lib/util/httpc.ex
@@ -53,6 +53,7 @@ defmodule Antikythera.Httpc do
   - `:skip_body_decompression` - By default gzip-compressed body is automatically decompressed (i.e. defaults to `false`).
     Pass `skip_body_decompression: true` if compressed body is what you need.
   - `:connect_options` - TCP/IP options supported by the `gen_tcp` erlang module.
+    Currently, only IP address family is supported.
   """
 
   require AntikytheraCore.Logger, as: L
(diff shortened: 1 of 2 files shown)


lib/util/httpc.ex
```diff
@@ -364,11 +366,18 @@ defmodule Antikythera.Httpc do
...
+    :connect_options, connect_options ->
+      {:connect_options, Enum.filter(connect_options, &filter_hackney_connect_option/1)}
+
     # :skip_body_decompression is used in processing response body, not here
     _, _ ->
       nil
   end
 
+  defunpt filter_hackney_connect_option(option :: :gen_tcp.connect_option()) :: v[boolean] do
+    option in [:inet, :inet6, :local]
```
aYukiSekiguchi, line 378: Do we need to supprt `:local`?

`:local` is [Unix Domain Socket](https://www.erlang.org/doc/man/gen_tcp.html#connect-4) . We need to pass `http+unix://` to Hackney to connect to a unix domain socket. However, `Httpc` only support `http(s)://` because `Httpc` uses `Url.t()` for URL and `Url.t()` [supports only the schemes](https://github.com/access-company/antikythera/blob/master/lib/type/string.ex#L106)

aYukiSekiguchi (changes requested):

lib/util/httpc.ex
irisTa56: Thank you.
For `Httpc`, only `:inet` and `:inet6` should be accepted.
Fixed at https://github.com/access-company/antikythera/pull/260/commits/8cd5f5154fadbbabc701b0a845d4149664cc6c07

aYukiSekiguchi (approved):
LGTM

PR #263, opened 2024-03-26
[179864] Avoid negated conditions in if else blocks

irisTa56:
https://acsmine.tok.access-company.com/redmine/issues/179864#note-29

### DIFF
--- a/core/version/gear.ex
+++ b/core/version/gear.ex
@@ -62,7 +62,7 @@ defmodule AntikytheraCore.Version.Gear do
     end)
   end
 
-  defunpt auto_generated_module?(mod :: atom) :: boolean do
+  defunpt auto_generated_module?(mod :: v[module]) :: v[boolean] do
     Atom.to_string(mod) |> String.starts_with?("Elixir.Croma.TypeGen.")
   end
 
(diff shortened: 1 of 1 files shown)


core/version/gear.ex
```diff
@@ -70,10 +62,22 @@ defmodule AntikytheraCore.Version.Gear do
     end)
   end
 
-  defunpt auto_generated_module?(mod :: atom) :: boolean do
+  defunpt auto_generated_module?(mod :: v[module]) :: v[boolean] do
     Atom.to_string(mod) |> String.starts_with?("Elixir.Croma.TypeGen.")
   end
 
+  defunp load_module_if_needed!(mod :: v[module]) :: :ok | nil do
```
aYukiSekiguchi, line 69: I understand you want to keep the return value of the original code. However, if a new member sees these return values, it's difficult to understand what `nil` means. I think this function should always return `:ok`.

aYukiSekiguchi (approved):
LGTM

core/version/gear.ex
irisTa56: Thank you.
It's better to always return `:ok` for success cases.
Fixed at https://github.com/access-company/antikythera/pull/263/commits/b2196802d27f4d619aa539c070485166b4f4a1c5

PR #304, opened 2025-06-18
[308633] Doc: Fix description of per-API timeout

Double-oxygeN:
https://acsmine.tok.access-company.com/redmine/issues/308633

### DIFF
--- a/lib/web/router/router.ex
+++ b/lib/web/router/router.ex
@@ -132,8 +132,10 @@ defmodule Antikythera.Router do
 
       get "/foo", Hello, :long_action, timeout: 60_000
 
+  If not set, the timeout is considered as 10 seconds.
+  This default value can be overwritten by `GEAR_ACTION_TIMEOUT` environment variable.
+
   The maximum timeout is determined by `:gear_action_max_timeout` configuration in `config/config.exs`.
-  The default value is 10 seconds, which can be configured by `GEAR_ACTION_TIMEOUT` environment variable.
   """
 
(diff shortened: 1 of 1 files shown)


lib/web/router/router.ex
```diff
@@ -132,8 +132,10 @@ defmodule Antikythera.Router do
 
       get "/foo", Hello, :long_action, timeout: 60_000
 
+  If not set, the timeout is considered as 10 seconds.
+  This default value can be overwritten by `GEAR_ACTION_TIMEOUT` environment variable.
```
aYukiSekiguchi, line 136: The previous documentation was a bit complicated because it tried to explain both the default value of `:timeout` and the default value of `GEAR_ACTION_TIMEOUT`. Since `GEAR_ACTION_TIMEOUT` is already documented in `development_environment.md`, I don’t think we need to repeat that information here.

To make this section clearer, I focused only on describing the `:timeout` option itself. I updated the documentation accordingly and asked Copilot to help improve the wording. Here’s the revised version:

```
  You can set a timeout for each API using the `:timeout` option, specified in milliseconds.
  For example, the following API times out after 60 seconds:

      get "/foo", Hello, :long_action, timeout: 60_000

  If not specified, the default timeout is taken from the `GEAR_ACTION_TIMEOUT` environment variable.
  The maximum allowed value is set by the `:gear_action_max_timeout` configuration in `config/config.exs`.
```

aYukiSekiguchi (approved):
LGTM

PR #305, opened 2025-06-19
[302690] Write gear log in the order of micoseconds

aYukiSekiguchi:
https://acsmine.tok.access-company.com/redmine/issues/302690


Double-oxygeN (approved):
LGTM

PR #333, opened 2026-04-20
[322048] Add Antikythera.Test.InProcessClient for in-process gear controller testing

aYukiSekiguchi:
https://acsmine.tok.access-company.com/redmine/issues/322048

### DIFF
--- a/core/handler/gear_action/web.ex
+++ b/core/handler/gear_action/web.ex
@@ -12,6 +12,7 @@ defmodule AntikytheraCore.Handler.GearAction.Web do
   alias AntikytheraCore.Handler.GearError
   alias AntikytheraCore.Handler.HelperModules
   alias AntikytheraCore.Handler.CowboyReq
+  alias AntikytheraCore.Handler.BodyParser
   alias AntikytheraCore.Handler.ExecutorPoolHelper
   alias AntikytheraCore.Handler.WebsocketState
   alias AntikytheraCore.Conn, as: CoreConn
@@ -69,13 +70,9 @@ defmodule AntikytheraCore.Handler.GearAction.Web do
     end
   end
 
-  defunp find_route(
-           req :: :cowboy_req.req(),
-           gear_name :: v[GearName.t()],
-           method :: v[Method.t()],
-           path_info :: v[PathInfo.t()],
-           %HelperModules{router: router}
-         ) :: R.t({GearEntryPoint.t(), PathMatches.t(), boolean, boolean, GearActionTimeout.t()}) do
+  defunp lookup_route(router :: v[module], method :: v[Method.t()], path_info :: v[PathInfo.t()]) ::
+           {:ok, {GearEntryPoint.t(), PathMatches.t(), boolean, boolean, GearActionTimeout.t()}}
+           | :no_route do
     case router.__web_route__(method, path_info) do
       # New format (6-tuple) with http_streaming flag
       {controller, action, path_matches, websocket?, http_streaming?, timeout} ->
@@ -86,6 +83,22 @@ defmodule AntikytheraCore.Handler.GearAction.Web do
         {:ok, {{controller, action}, path_matches, websocket?, false, timeout}}
 
       nil ->
+        :no_route
+    end
+  end
+
+  defunp find_route(
+           req :: :cowboy_req.req(),
+           gear_name :: v[GearName.t()],
+           method :: v[Method.t()],
+           path_info :: v[PathInfo.t()],
+           %HelperModules{router: router}
+         ) :: R.t({GearEntryPoint.t(), PathMatches.t(), boolean, boolean, GearActionTimeout.t()}) do
+    case lookup_route(router, method, path_info) do
+      {:ok, result} ->
+        {:ok, result}
+
+      :no_route ->
         {:error,
          CowboyReq.with_conn(
            req,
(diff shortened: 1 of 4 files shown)


lib/test/in_process_client.ex
```diff
@@ -0,0 +1,295 @@
...
+
+    option_headers =
+      %{}
+      |> add_cookie_header(Keyword.get(options, :cookie))
+      |> add_basic_auth_header(Keyword.get(options, :basic_auth))
+
+    req_headers =
+      body_headers
+      |> Map.merge(option_headers)
+      |> Map.merge(downcase_keys(headers))
```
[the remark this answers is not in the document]
aYukiSekiguchi: Fixed in 5cb084c — added `Map.put_new("accept-encoding", "gzip")` to match Httpc's default.

core/handler/gear_action/web.ex
```diff
@@ -405,4 +418,70 @@ defmodule AntikytheraCore.Handler.GearAction.Web do
...
+              GearAction.Context.make(),
+              helper_modules,
+              fn ->
+                case ActionRunner.run_action(conn, entry_point) do
+                  {:ok, conn2} -> CoreConn.run_before_send(conn2, conn)
+                  {:error, reason, st} -> GearError.error(conn, reason, st)
+                end
+              end
+            )
+
```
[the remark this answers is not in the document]
aYukiSekiguchi: Fixed in 06135b0 — `handle_in_process` now handles `{:error, _}` from `CowboyReq.query_params/2` by returning `GearError.bad_request/1`, matching the production 400 response. A `safe_query_params/2` helper covers the already-bad-request branch with empty params.

lib/test/in_process_client.ex
```diff
@@ -0,0 +1,295 @@
...
+    fake_req = %{
+      method: method |> Atom.to_string() |> String.upcase(),
+      version: :"HTTP/1.1",
+      scheme: uri.scheme || "http",
+      host: uri.host || "localhost",
+      port: uri.port || 80,
+      path: path_only,
+      path_info: decoded_path_info,
+      qs: qs,
+      headers: req_headers,
```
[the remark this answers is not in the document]
aYukiSekiguchi: Fixed in 39a3b1c — added a default `host` header via `Map.put_new("host", default_host_header(scheme, host, port))`. The helper omits the port for default http/80 and https/443, matching typical HTTP client behavior.

lib/test/in_process_client.ex
```diff
@@ -0,0 +1,295 @@
...
+
+        body when is_binary(body) and byte_size(body) > 0 ->
+          {body, %{"content-length" => byte_size(body) |> Integer.to_string()}}
+
+        body when is_binary(body) ->
+          {"", %{}}
+
+        body when is_map(body) or is_list(body) ->
+          raw = Poison.encode!(body)
+          {raw, content_headers(raw, "application/json")}
```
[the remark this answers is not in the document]
aYukiSekiguchi: Fixed in 01e5abe — added a `{:file, path}` clause using `File.read!/1`, and split the previous `is_map(body) or is_list(body)` clause: maps are still JSON-encoded, while lists are treated as iodata via `IO.iodata_to_binary/1` (matching hackney's behavior). Extracted a `raw_body_with_length/1` helper to share the common length-header construction.

lib/test/in_process_client.ex
```diff
@@ -0,0 +1,321 @@
...
+    parts |> Enum.reject(&is_nil/1) |> Enum.join("&")
+  end
+
+  defunp downcase_keys(map :: v[Headers.t()]) :: v[Headers.t()] do
+    Map.new(map, fn {k, v} -> {String.downcase(k), v} end)
+  end
+
+  # Build a default HTTP/1.1 `Host` header. Include the port only when it is
+  # non-default so the value matches what a typical HTTP client would send.
+  defunp default_host_header(scheme :: v[String.t()], host :: v[String.t()], port :: v[non_neg_integer]) :: v[String.t()] do
```
Double-oxygeN: [nitpick] Ran `mix format`.

```suggestion
  defunp default_host_header(
           scheme :: v[String.t()],
           host :: v[String.t()],
           port :: v[non_neg_integer]
         ) :: v[String.t()] do
```

lib/test/in_process_client.ex
```diff
@@ -0,0 +1,321 @@
...
+  }
+
+  defun dispatch(
+          gear_name :: v[GearName.t()],
+          method :: v[Method.t()],
+          url :: v[String.t()],
+          body :: ReqBody.t(),
+          headers :: v[Headers.t()],
+          options :: Keyword.t()
+        ) :: v[Response.t()] do
```
Double-oxygeN, line 119: Is this function also intended to be called directly from gears without going through macro expansion? If not, I recommend adding `@doc false`.

lib/test/in_process_client.ex
aYukiSekiguchi: Fixed in 38e090b — applied `mix format` to `lib/test/in_process_client.ex`.

lib/test/in_process_client.ex
aYukiSekiguchi: Fixed in 4ef59aa — added `@doc false` to `dispatch/6`. The intended public API is the `use Antikythera.Test.InProcessClient` macro, which binds `gear_name` implicitly and exposes the `get`/`post`/etc. helpers. Direct calls to `dispatch/6` are an internal implementation detail.

aYukiSekiguchi: Additional context: I originally considered exposing `dispatch/6` so gear developers could replace direct `Httpc.request/5` calls, even though `HttpClient` doesn't have an equivalent API. However, after searching the codebase I found that only a few tests use `Httpc.request/5` in that way. So keeping `dispatch/6` as internal for now seems fine — we can add a `request/5`-compatible public API later if there's demand.

Double-oxygeN (approved):
LGTM

PR #339, opened 2026-06-12
[323509] Add mimic 1.7.4 as a test dependency

aYukiSekiguchi:
https://acsmine.tok.access-company.com/redmine/issues/323509


Double-oxygeN (approved):
LGTM

(optional) It would be helpful to include an explanation for users as to why there is a dependency on both the meck and mimic libraries.

aYukiSekiguchi:
Thanks for the review! I added a comment in `mix.exs` explaining why both `meck` and `mimic` are kept as test dependencies ( 34a9fd0 ).

